PR #101, opened 2017-08-01
Rest-Client: add method to change the base uri of the client

sebastianvogel:


rest-client/src/main/java/com/gentics/mesh/rest/client/AbstractMeshRestHttpClient.java
```diff
@@ -28,6 +27,8 @@
 
 	protected boolean disableAnonymousAccess = false;
 
+	private String baseUri;
```
Jotschi: Keep the default value of "/api/v1/". Don't handle the default value in the getter.

rest-client/src/main/java/com/gentics/mesh/rest/client/AbstractMeshRestHttpClient.java
```diff
@@ -171,4 +168,22 @@ public static String getQuery(ParameterProvider... parameters) {
...
+	 * @param baseUri
+	 * 				the base URI
+	 */
+	public void setBaseUri(String baseUri) {
+		this.baseUri = baseUri;
+	}
+
+	@Override
+	public String getBaseUri() {
+		if (baseUri == null) {
```
Jotschi: Remove this

PR #151, opened 2017-09-25
Improve performance of removing permissions

philippguertler:


core/src/main/java/com/gentics/mesh/core/data/impl/RoleImpl.java
```diff
@@ -122,10 +124,18 @@ private void setGroups(InternalActionContext ac, RoleResponse restRole) {
 
 	@Override
 	public void revokePermissions(MeshVertex node, GraphPermission... permissions) {
```
Jotschi: Rename node to vertex

Issue #163, opened 2017-10-13
GraphQL: Add referencedBy node field to Node type


Jotschi:
Add a new field to the node type which allows the user of the API to load referenced nodes.

(nodeA:Node)-[:HAS_CONTAINER]->(NodeContainer)-[:HAS_NODE_FIELD]-(nodeB:Node)

The added field should be able to find `nodeA` when using `nodeB`

Node list fields may have different relational edges and must be handled separately. Both relations must be merged into a single traversal in order to retrieve a correct pages resultset.

The referenced nodes must also be checked for release and version. (We only want to find published/draft nodes)

```json
{
  node(uuid: "21203632520b4d19a03632520b2d19c1") {
      foreignReferences(fieldName: "imageNodes", schemaName: "content") {
      // (Pages) Lists all content nodes which have imageNodes node field which references the field
      // Find all node containers (in the expected version) which reference this node via a node field  
    }
    foreignReferences {
      // (Paged) Lists all nodes which reference the current node via any node field
      node
      fieldName
    }
  }
}
```

PR #178, opened 2017-10-27
GraphQL - referencedBy enhancement

philippguertler:
Resolves https://github.com/gentics/mesh-incubator/issues/116


Jotschi:
#163

philippguertler:
Diff too large, this is now https://github.com/gentics/mesh/pull/773

PR #193, opened 2017-11-10
Updated microschema deletion test

### LINKED ISSUE #192: Unable to delete Microschemas
- v0.10.4

## Operating System
- Linux
- MacOSX
- Windows

## JVM
- Open JDK 1.8.0_151

## Problem

There's a prevention from deleting a microschema if it is still in use in a node. However, this seems like it's not updating properly, as when I delete all nodes with such a microschema, I still get an error when attempting to delete it:

```json
{
	"i18nParameters": [
		"8c5cee4596f744469cee4596f7244689"
	],
	"properties": {
	},
	"type": "generic_error",
	"message": "There are micronodes which use the microschema 8c5cee4596f744469cee4596f7244689. Delete the nodes which use the micronodes first."
}
```

I also noticed that sometimes a restart can help, but not always.

## Reproducer

- Create a Microschema and a Schema which has a reference to the new microschema
- Create a new node with the newly created schema
- Delete the node
- Attempt to delete the microschema
- You get an error as described above

That it's possible to delete it without having to restart the server

SiNEXiST:


Jotschi:
@SiNEXiST Okay thanks - I'll take a look at it.

PS: Don't add the "Fixes xyz" line to that commit. That would close the issue and the change in this PR is not actually fixing the issue ;)

PR #276, opened 2018-02-05
Dev 0.15.x schema fields

yrucrem:
Make schema fields available in GraphQL

### DIFF
--- a/core/src/test/resources/graphql/schema-query
+++ b/core/src/test/resources/graphql/schema-query
@@ -70,7 +70,34 @@
 		        }
 		    }
 		}
+
+		# [$.data.schema.fields[0].name=slug]
+		# [$.data.schema.fields[0].label=Slug]
+		# [$.data.schema.fields[0].required=true]
+		# [$.data.schema.fields[0].type=string]
+		# [$.data.schema.fields[1].name=title]
+		# [$.data.schema.fields[1].label=Title]
+		# [$.data.schema.fields[1].required=false]
+		# [$.data.schema.fields[1].type=string]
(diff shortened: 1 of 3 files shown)


core/src/test/resources/graphql/schema-query
```diff
@@ -70,7 +70,34 @@
...
+		# [$.data.schema.fields[0].type=string]
+		# [$.data.schema.fields[1].name=title]
+		# [$.data.schema.fields[1].label=Title]
+		# [$.data.schema.fields[1].required=false]
+		# [$.data.schema.fields[1].type=string]
+		# [$.data.schema.fields[3].name=content]
+		# [$.data.schema.fields[3].label=Content]
+		# [$.data.schema.fields[3].required=false]
+		# [$.data.schema.fields[3].type=html]
+		fields {
```
Jotschi: Please also assert the array length

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/SchemaTypeProvider.java
```diff
@@ -95,10 +97,21 @@ public GraphQLObjectType createType() {
 				return new NodeContent(node, container);
 			});
 
-			return new DynamicStreamPageImpl(nodes, getPagingInfo(env));
+			return new DynamicStreamPageImpl<>(nodes, getPagingInfo(env));
 		}, NODE_PAGE_TYPE_NAME));
 
-		// TODO add fields
+		String SCHEMA_FIELD_TYPE = "SchemaFieldType";
```
Jotschi: Move to top of class & make it final public static

PR #284, opened 2018-02-13
Implement checks for allow field in schema

philippguertler:
Relates to #74

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/node/field/list/impl/NodeGraphFieldListImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/node/field/list/impl/NodeGraphFieldListImpl.java
@@ -2,6 +2,7 @@ package com.gentics.mesh.core.data.node.field.list.impl;
 
 import static com.gentics.mesh.core.rest.error.Errors.error;
 import static io.netty.handler.codec.http.HttpResponseStatus.BAD_REQUEST;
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/gentics/mesh/core/data/node/field/list/impl/NodeGraphFieldListImpl.java
```diff
@@ -81,6 +82,9 @@
 			if (node == null) {
 				throw error(BAD_REQUEST, "node_list_item_not_found", item.getUuid());
 			}
+			if (!fieldSchema.isAllowedSchema(node.getSchemaContainer().getName())) {
+				throw error(NOT_FOUND, "node_error_reference_not_allowed_schema", item.getUuid(), fieldKey);
```
Jotschi: Add the info from which node the error was caused. This would be very handy when dealing with batch updates or node migration errors.

philippguertler:
This change could cause problems with existing implementations. Here is an example:
1. There is a node which has a node field with the allow schema restriction set.
2. In this node there is a reference to a node which has a schema that is **not** allowed.
3. This update is applied.
4. The schema of the node is updated.
The schema migration will most likely fail because the schema migration checks now also check the allow setting. The Mesh instance is now in a broken state because the migration cannot be reverted.

This change should only be merged after reverting schema migrations is possible.

PR #289, opened 2018-02-13
Dedicated database revision hash

Jotschi:


Jotschi:
TODO: Add a way to document the database revision per Gentics Mesh version

PR #300, opened 2018-02-19
NPE when keystore path contains no directory

elbird:
NPE on startup when keystore path contains no directory eg: keystorePath: "keystore.jceks"


core/src/main/java/com/gentics/mesh/cli/MeshImpl.java
```diff
@@ -123,11 +123,20 @@ public void run(boolean block) throws Exception {
 
 	private void setupKeystore(MeshOptions options) throws Exception {
 		String keyStorePath = options.getAuthenticationOptions().getKeystorePath();
+		if (keyStorePath == null || keyStorePath.isEmpty()) {
```
Jotschi: Move this into the validation section of the MeshOptions / AuthenticationOptions.

elbird: Done

PR #361, opened 2018-03-26
Fix rest client never returns on webroot request containing spaces

elbird:


Jotschi:
@elbird You were right. URIUtils#encodeFragment used the wrong method although the added checks corrected the encoding process. I renamed and updated the method in the dev branch. I'll resolve the conflict once the review is done. URIUtils is tested.

Jotschi:
@philippguertler Please take a look at the code. LGTM

PR #370, opened 2018-04-05
Fixed possible Null-Pointer in mapping generation

SiNEXiST:


elasticsearch/src/main/java/com/gentics/mesh/search/index/node/NodeContainerMappingProvider.java
```diff
@@ -275,6 +275,11 @@ public JsonObject getFieldMapping(FieldSchema fieldSchema, Release release) {
 	 * @return An Properties-mapping for a microschema field.
 	 */
 	public JsonObject getMicroschemaMappingOptions(String[] allowed, Release release) {
```
Jotschi: Would it be possible to use (Release release, String... allowed) instead?

SiNEXiST: The release is in most cases not given, while the allowed microschemas are. I don't mind changing it, but then I'd propose:
```
getMicroschemaMappingOptions(String...allowed) {
return getMicroschemaMappingOptions(null, allowed);
}

getMicroschemaMappingOptions(Release release, String... allowed) {
// ...
}
```

PR #373, opened 2018-04-05
Fix creation of release with uuid: enqueue node migration

npomaroli:


core/src/main/java/com/gentics/mesh/core/data/impl/ReleaseImpl.java
```diff
@@ -435,4 +437,10 @@ public void delete(SearchQueueBatch batch) {
 		getVertex().remove();
 	}
 
+	@Override
+	public void onCreated() {
+		super.onCreated();
+		MeshInternal.get().boot().jobRoot().enqueueReleaseMigration(getCreator(), this);
```
Jotschi: I moved this into ReleaseRoot#create since it is better to handle job creation inside the tx which must be committed. The async tx in HandlerUtils#createOrUpdateElement should only be used to read.

PR #380, opened 2018-04-12
Fix error on node without content

philippguertler:


Jotschi:
It would be better to fix the inconsistency which caused the issue. I don't see how this situation happen during normal operation.

philippguertler:
Empty content can also be returned when there is no language of a certain node. I will add tests for these cases.

PR #427, opened 2018-05-29
Rename releases into branches

Jotschi:


Jotschi:
* [ ] Rename edge properties
* [ ] Rename release edges
* [ ] Rename vertex types (e.g. drop old type)
* [ ] Check which query parameter changed
* [ ] Check which REST endpoint paths changed
* [ ] Check whether REST properties changed

PR #436, opened 2018-06-12
OrientDB 3.0.x

Jotschi:


mephinet:
Will not be merged, because we already have a better implementation. Left here as a reference.

PR #444, opened 2018-06-20
Force UTF-8 encoding for static admin ui resources

alshain:
On windows (default charset 1252), the admin UI is shown with mangled umlauts. This should instruct vertx to decode the resources as UTF-8 no matter the default locale.

image


Jotschi:
Reproduced. Fix validated. I'll add the missing changelog entry. Thanks for the contribution!

alshain:
@Jotschi Wow, that was quick! No problem and thanks a lot.

PR #451, opened 2018-06-22
Change breadcrumb output. Fixes #398

### LINKED ISSUE #398: Breadcrumb Enhancements
* Reverse the order of the breadcrumb
* Include the basenode in the breadcrumb - https://github.com/gentics/mesh/blob/19f402a82025ae94dacaafe23053378247d70cba/core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java#L861

philippguertler:


philippguertler:
Added test for root node breadcrumb and updated docs. I have tested the change with the current UI and everything seems to work. I have also searched the code and it seems that the UI does not rely on the breadcrumb property at all.

PR #470, opened 2018-07-12
Fix order of elements in micronode list. Fixes #469

### LINKED ISSUE #469: Order of elements in a field of type list of micronodes is not preserved (when 40 or more elements)
- v0.24.4

## Operating System

- MacOSX

## Problem

A list of micronode in a node is shuffled when 40 or more elements in it:

## Reproducer

Create a simple node schema with 1 field pointing to a LIST of microschema (with for eg one text/number field)

create 39 element in the list
save --> order is preserved

image

add 1 element (40 element in the list)
image

save --> order is lost
image

after some debug in local i see that the micronode list is still in the same order when saving it
but when retreiving it to rest model it's wrong order
perhaps addItem in graph with ferma is bugged after 40 items

order must be preserved

Jotschi:


Jotschi:
TODO:
* [ ] Changelog

cschockaert:
Hello
Seems fine to me (as a quick dirty workaround), this branch will be merged into 0.25.X version?

This meens that addFramedEdge isn't working well with more than 39 elements?

Jotschi:
@cschockaert Yes. It will be hopefully included in __0.21.5__ - No `addFramedEdge` works as expected. Sorting is in general required if deal with a Graph like this:



The `post` vertices are just connected to the `likes` vertex. If you now load all `post` vertices you will need to sort to ensure the order. Otherwise the order can't be ensured due to the nature of the graph and the way the GraphDB stored new edges.

PR #507, opened 2018-10-02
Add GraphQL filtering for users, groups and roles

philippguertler:
Also updates graphql-java to 10.0 and GraphiQL to 0.12.0


core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -1471,26 +1469,19 @@ public void deleteFromBranch(InternalActionContext ac, Branch branch, BulkAction
...
-						continue;
-					}
-					String languageTag = edge.getProperty(GraphFieldContainerEdgeImpl.LANGUAGE_TAG_KEY);
-					if (languageTags.contains(languageTag)) {
+		if (languageTags == null) {
+			return new DynamicTransformablePageImpl<>(ac.getUser(), indexName, indexKey, NodeImpl.class, pagingInfo, perm, null, true);
+		} else {
+			return new DynamicTransformablePageImpl<NodeImpl>(ac.getUser(), indexName, indexKey, NodeImpl.class, pagingInfo, perm, (item) -> {
+				// Filter out nodes which do not provide one of the specified language tags and type
+				for (NodeGraphFieldContainer container : item.getGraphFieldContainersIt(branchUuid, type)) {
```
Jotschi: Better to access the framed edge instead of the framed field container.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/CommonFields.java
```diff
@@ -0,0 +1,44 @@
...
+public final class CommonFields {
+	private CommonFields() {}
+
+	public static <T extends MeshElement> FilterField<T, Map<String, ?>> uuidFilter() {
+		return new MappedFilter<>("uuid", "Filters by uuid", StringFilter.filter(), MeshElement::getUuid);
+	}
+
+	public static <T extends NamedElement> List<FilterField<T, Map<String, ?>>> nameFilter() {
+		return Arrays.asList(
+			uuidFilter(),
```
Jotschi: Remove uuid filter here. It is better to let the implementer choose which filters to include.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/GroupFilter.java
```diff
@@ -0,0 +1,33 @@
+package com.gentics.mesh.graphql.filter;
+
+import com.gentics.graphqlfilter.filter.FilterField;
+import com.gentics.graphqlfilter.filter.StartMainFilter;
+import com.gentics.mesh.core.data.Group;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class GroupFilter extends StartMainFilter<Group> {
```
Jotschi: javadoc description

verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/NodeFilter.java
```diff
@@ -42,8 +42,12 @@ private NodeFilter(GraphQLContext context) {
 			.add(new MappedFilter<>("schema", "Filters by schema", SchemaFilter.filter(context), content -> content.getNode().getSchemaContainer()));
 		filters.add(new MappedFilter<>("created", "Filters by node creation timestamp", DateFilter.filter(),
 			content -> content.getNode().getCreationTimestamp()));
+		filters.add(new MappedFilter<>("creator", "Filters by creator", UserFilter.filter(),
+			content -> content.getNode().getCreator()));
 		filters.add(new MappedFilter<>("edited", "Filters by node update timestamp", DateFilter.filter(),
 			content -> content.getContainer().getLastEditedTimestamp()));
```
Jotschi: Check whether it may be possible that getContainer() could return null. Best to trace the getContainer calls and check whether any null handling is already in place somewhere in the code.

PR #509, opened 2018-10-04
Allow local plugin config file (config.local.yml) 

patrickhoefer:
to override parts of config.yml

### DIFF
--- a/plugin-api/src/main/java/com/gentics/mesh/plugin/AbstractPluginVerticle.java
+++ b/plugin-api/src/main/java/com/gentics/mesh/plugin/AbstractPluginVerticle.java
@@ -1,19 +1,13 @@
 package com.gentics.mesh.plugin;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
 
-import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
 
-import com.fasterxml.jackson.annotation.JsonInclude.Include;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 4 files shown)


plugin-api/src/main/java/com/gentics/mesh/plugin/AbstractPluginVerticle.java
```diff
@@ -93,17 +76,22 @@ public File getStorageDir() {
...
+		if (!configFile.canRead()) {
 			return null;
 		}
-		try (FileInputStream fis = new FileInputStream(configFile)) {
-			return getYAMLMapper().readValue(fis, clazz);
+		T config = PluginConfigLoader.loadConfig(configFile, null, clazz);
+
+		//try to load local config file
+		File localConfigFile = getLocalConfigFile();
+		if(localConfigFile.canRead()) {
```
Jotschi: Formatting

plugin-api/src/main/java/com/gentics/mesh/plugin/AbstractPluginVerticle.java
```diff
@@ -93,17 +76,22 @@ public File getStorageDir() {
...
+			config = PluginConfigLoader.loadConfig(localConfigFile, config, clazz);
 		}
+		return config;
 	}
 
 	@Override
 	public <T> T writeConfig(T config) throws IOException {
-		String yaml = getYAMLMapper().writeValueAsString(config);
```
Jotschi: Offload whole writeConfig to PluginConfigLoader - The class should also handle write (thus rename the class)

Jotschi:
I added getYAMLMapper back to the abstract class to avoid compile errors for older plugins:
```
	/**
	 * Use {@link PluginConfigUtil#getYAMLMapper()} instead.
	 * @deprecated
	 * @return
	 */
	@Deprecated
	public static ObjectMapper getYAMLMapper() {
		return PluginConfigUtil.getYAMLMapper();
	}
```

PR #511, opened 2018-10-05
Fix publish consistency check when moving folders

philippguertler:


Jotschi:
Check tests and merge

PR #526, opened 2018-10-30
Branch Path Prefix Handling

Jotschi:


common/src/main/java/com/gentics/mesh/core/link/WebRootLinkReplacer.java
```diff
@@ -217,9 +219,19 @@ public String resolve(InternalActionContext ac, String branchUuid, ContainerType
 		if (log.isDebugEnabled()) {
 			log.debug("Resolving link to " + node.getUuid() + " in language " + Arrays.toString(languageTags) + " with type " + type.name());
 		}
+
 		String path = node.getPath(ac, branchUuid, edgeType, languageTags);
 		if (path == null) {
 			path = "/error/404";
+		} else {
+			String prefix = branch.getPathPrefix();
```
npomaroli: if a branchUuid was passed to this method, branch will be null here

common/src/main/java/com/gentics/mesh/core/link/WebRootLinkReplacer.java
```diff
@@ -217,9 +219,19 @@ public String resolve(InternalActionContext ac, String branchUuid, ContainerType
...
+			if (prefix != null) {
+				// Ensure that the prefix starts with a slash
+				if (!prefix.startsWith("/")) {
```
npomaroli: What if the prefix ends with a slash or is an empty string?

core/src/main/java/com/gentics/mesh/core/data/impl/BranchImpl.java
```diff
@@ -113,6 +113,11 @@ public boolean update(InternalActionContext ac, SearchQueueBatch batch) {
 			modified = true;
 		}
 
+		if (shouldUpdate(requestModel.getPathPrefix(), getPathPrefix())) {
```
npomaroli: Does this mean, that it is impossible to set the pathPrefix to empty?

PR #531, opened 2018-11-05
Dev nio binary

Jotschi:


philippguertler:
This PR will disabled gzip compression for binary downloads.

Issue #538, opened 2018-11-09
Catch with no Log


sunsus:
Please log at least a Debug here, otherwiese you woll newer know if there was a connection problem: 
https://github.com/gentics/mesh/blob/1d8457026af04a9ef170f1b0f148765af3675222/elasticsearch/src/main/java/com/gentics/mesh/search/impl/ElasticSearchProvider.java#L148

Jotschi:
We actually hope to get rid of the whole loop in https://github.com/gentics/mesh/blob/1d8457026af04a9ef170f1b0f148765af3675222/elasticsearch/src/main/java/com/gentics/mesh/search/impl/ElasticSearchProvider.java#L133 - I'll update the error handling and log connection issues in the meanwhile.

Jotschi:
Gentics Mesh will no longer wait for ES. This change has been released with 0.32.0

PR #539, opened 2018-11-09
Catch with no Log

sunsus:
Fix:  #538


Jotschi:
@sunsus Thanks for the PR. I have added a similar log to the class - the change will be included in the next release.

PR #540, opened 2018-11-09
Make OAuth2 Server options overrideable via env vars

elbird:


elbird:
runnow

api/src/main/java/com/gentics/mesh/etc/config/env/OptionUtils.java
```diff
@@ -0,0 +1,90 @@
...
+
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
+public class OptionUtils {
```
Jotschi: Make this final
```suggestion
public final class OptionUtils {

	private OptionUtils() {
	}
```

PR #547, opened 2018-11-14
Fix /migrateSchemas endpoint

### LINKED ISSUE #532: Node Migration Race Condition
## Situation

1. Create Node
2. Update schema (with failing migration)
3. Update schema again (no error since no other node needs migration)
4. Invoke `/api/v1/:project/branches/:branchUuid/migrateSchemas`

or

1. Create Node
2. Update Schema (with failing migration)
3. Unassign schema from project
4. Invoke `/api/v1/:project/branches/:branchUuid/migrateSchemas`

## Issue

An error should be observed because the preparation would create indices for older schema versions which the `NodeIndexHandler` rejects (since it deems them to no longer be active).

## Solutions

* Ignore `NodeIndexHandler` checks and create the index.
* Update `NodeIndexHandler` to also check for older not migrated nodes (could be a performance issue since all NGFC or SchemaVersions would need to be checked)

philippguertler:
Fixes #532

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -40,6 +40,8 @@ icon:check[] REST: A concurrency issue has been fixed which could lead to proble
 
 icon:check[] REST: Fix error message when no Node for a GraphFieldContainer can be found. link:https://github.com/gentics/mesh/issues/364[#364]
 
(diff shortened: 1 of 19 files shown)


philippguertler:
Change it for microschemas as well. Please review again.

Jotschi:
LGTM - Run tests

PR #551, opened 2018-11-15
Migrate nodes to newest schema when creating new branch

### LINKED ISSUE #521: Nodes get migrated to new branch in old schema version
- v0.26.0

## Problem

When a new branch is created, it automatically gets the newest versions of the project's schemas assigned, but node will get migrated with schema versions used in base branch

## Reproducer

Create schema (version 1.0)
Create branch (gets schema version 1.0 assigned)
Create node in branch (using schema version 1.0)
Update schema without assigning to branches (version 2.0)
Create new branch (gets schema version 2.0 assigned)

Expected behaviour: node should use schema version 2.0 after migration to new branch
Actual behaviour: node is migrated to new branch, but still uses schema version 1.0

philippguertler:
Fixes #521

### DIFF
--- a/core/src/test/java/com/gentics/mesh/core/branch/BranchEndpointTest.java
+++ b/core/src/test/java/com/gentics/mesh/core/branch/BranchEndpointTest.java
@@ -81,7 +81,6 @@ import com.gentics.mesh.test.util.TestUtils;
 import com.gentics.mesh.util.UUIDUtil;
 import com.syncleus.ferma.tx.Tx;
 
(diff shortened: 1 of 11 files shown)


core/src/test/java/com/gentics/mesh/core/branch/BranchEndpointTest.java
```diff
@@ -75,6 +80,7 @@
 import com.gentics.mesh.util.UUIDUtil;
 import com.syncleus.ferma.tx.Tx;
 
+//@MeshTestSetting(useElasticsearch = false, testSize = FULL, startServer = true, inMemoryDB = false, startStorageServer = true)
```
Jotschi:
```suggestion
```

PR #560, opened 2018-11-16
Partial refactor to Tinkerpop 3.x compliant API / MADL support

Jotschi:


Jotschi:
TODO: Check for sideeffects during element removal

PR #572, opened 2018-11-20
Test Fixes

Jotschi:


Jotschi:
TODO:
* [ ] Add tests for micronode lists as well
* [ ] Check for additional complex fields
* [ ] Check whether deleting last (the only) language will result in the same clean result
* [ ] Add consistency checks for this situation
* [ ] Check how dangling micronodes can be removed

PR #577, opened 2018-11-21
Add parallel upload test with same binary

Jotschi:


Jotschi:
Fix already included in dev - Test has been cherry-picked.

PR #581, opened 2018-11-22
Add upsertNode method to Java REST client

Jotschi:


rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
```diff
@@ -143,6 +144,15 @@ public MeshRestClient disableAnonymousAccess() {
 		return prepareRequest(GET, "/" + encodeSegment(projectName) + "/nodes/" + uuid + getQuery(parameters), NodeResponse.class);
 	}
 
+	@Override
+	public MeshRequest<NodeResponse> upsertNode(String projectName, String uuid, NodeUpsertRequest nodeUpsertRequest,
```
Jotschi: I know that the signature is different from `createNode(uuid, projectName)` but I prefer this order since it also reflects the ids in the REST path

Jotschi:
* Add upsert for other types as well
* Add upsert via webroot

rest-model/src/main/java/com/gentics/mesh/core/rest/node/NodeUpdateRequest.java
```diff
@@ -20,9 +20,9 @@
 	@JsonPropertyDescription("Dynamic map with fields of the node content.")
 	private FieldMap fields = new FieldMapImpl();
 
-	@JsonProperty(required = true)
-	@JsonPropertyDescription("Version number which must be provided in order to handle and detect concurrent changes to the node content.")
-	private String version;
+	@JsonProperty(required = false)
+	@JsonPropertyDescription("Version number which can be provided in order to handle and detect concurrent changes to the node content.")
```
Jotschi: Don't handle the default in the POJO. Instead update the value within the REST handler.

PR #583, opened 2018-11-23
Don't detect changes on empty properties

### LINKED ISSUE #165: Change schema update request defaults
- v0.10.4

## Problem
When updating a schema and not providing the `container` property, the property is always set to false.

## Expected behaviour
Expected the container property to stay the same as before the update.

### LINKED ISSUE #496: Unset properties causes schema migration
- v0.22.1

## Problem

We use a script to upload all local schema files to the server to keep it in sync.
In the local schema, we don't have a `description` property set however, which causes the `schemas/{uuid}/diff` endpoint and subsequently the schema-migration to register changes:

```json
{
	"changes": [
		{
			"operation": "UPDATESCHEMA",
			"properties": {
				"description": null
			}
		}
	]
}
```

So everytime we upload the schema, it'll detect a "change" and perform a new schema migration.

## Reproducer

- Create a Schema without a description field
- Update the schema with the exact same content

It should detect that there is actually no change and not trigger unneccessary schema migrations.

philippguertler:
Fixes #496 
Fixes #165

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -36,6 +36,8 @@ icon:check[] Core: Nodes are now migrated to the newest schema version when a ne
 
 icon:check[] Core: Fixed a bug that prevented moving a published node in one language to a published container of another language.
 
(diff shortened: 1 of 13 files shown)


philippguertler:
Issues have already been fixed.

PR #586, opened 2018-11-25
Webroot index fix

Jotschi:


Jotschi:
Index works as expected - No need to change it.

PR #592, opened 2018-11-29
Schema create with UUID

Jotschi:


Jotschi:
Has now been implemented differently.

PR #604, opened 2019-01-08
Add okhttp rest client

philippguertler:


rest-client/pom.xml
```diff
@@ -33,6 +33,16 @@
 			<groupId>io.vertx</groupId>
 			<artifactId>vertx-rx-java2</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.squareup.okhttp3</groupId>
+			<artifactId>okhttp</artifactId>
+			<version>3.12.1</version>
```
Jotschi: Not a must have but I tend to use properties for central dependency versions. That makes it easier to update the whole stack.

Jotschi:
LGTM so far.  We need to check SSL support as well (not sure whether we have dedicated tests for this).

philippguertler:
I wrote a small test that connects to https://demo.getmesh.io and loads all projects, which works.

PR #611, opened 2019-01-14
Make sure images keep their file format after manipulations

yrucrem:
When no writer plugin for the image type is available the fallback is PNG.

SUP-7297


services/image-imgscalr/src/main/java/com/gentics/mesh/image/ImgscalrImageManipulator.java
```diff
@@ -134,6 +137,105 @@ protected BufferedImage resizeIfRequested(BufferedImage originalImage, ImageMani
...
+	 * Create an image reader for the given input.
+	 *
+	 * @param input The input stream to read the original image from.
+	 * @return An image reader reading from the given input stream
+	 */
+	private ImageReader getImageReader(ImageInputStream input) {
+		Iterator<ImageReader> it = ImageIO.getImageReaders(input);
+
+		if (!it.hasNext()) {
+			// No reader available for this image type.
```
Jotschi: Log a meaningful error. Otherwise we can't figure out whats wrong.

PR #626, opened 2019-01-21
Replace Vert.x client with OkHttpClient

philippguertler:


rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestOkHttpClientImpl.java
```diff
@@ -77,6 +79,11 @@ private static OkHttpClient createClient() {
 	}
 
 	private String getUrl(String path) {
-		return origin + getBaseUri() + path;
+		return config.getBaseUrl() + path;
+	}
+
+	@Override
+	public void close() {
+		// TODO close default client?
```
philippguertler: Explain why we don't close

PR #631, opened 2019-01-24
Dedicated handling for restore in clustered mode

Jotschi:


core/src/test/java/com/gentics/mesh/core/admin/AdminEndpointBackupTest.java
```diff
@@ -47,6 +49,12 @@ public void testBackupRestore() throws IOException {
 		call(() -> client().findNodeByUuid(NEW_PROJECT_NAME, baseNodeUuid), NOT_FOUND, "project_not_found", NEW_PROJECT_NAME);
 	}
 
+	@Test
+	public void testRestoreInClustereMode() {
```
philippguertler:
```suggestion
	public void testRestoreInClusterMode() {
```

PR #641, opened 2019-01-31
Add fixes to reduce cluster contention and ignore index unknown error

Jotschi:


elasticsearch/src/main/java/com/gentics/mesh/search/index/entry/AbstractIndexHandler.java
```diff
@@ -340,7 +340,8 @@ public Completable createIndex(CreateIndexEntry entry) {
 					log.debug("Index name {" + idx + "}");
 				}
 			}
-			throw error(INTERNAL_SERVER_ERROR, "error_index_unknown", indexName);
+			log.warn("Entry references an unknown index: {}", indexName);
+			return Completable.complete();
```
philippguertler: action.auto_create_index should be false, or else elasticsearch will create the index by itself

PR #642, opened 2019-01-31
Remove mesh version from node name

Jotschi:
Adding the mesh version to the node name causes orientdb to add bogus entries to its distributed-config.json file which lists all known nodes.


cschockaert:
thanks, ill not have to template and delete this file anymore when i upgrade the mesh version

PR #655, opened 2019-02-25
S3 Support / Binary Plugins

Jotschi:
This PR aims to provide S3 and additional binary plugin support

* BinaryAuxSourcePlugin - gentics/mesh-incubator#205
* BinaryProcessorPlugin - gentics/mesh-incubator#27
* BinaryStoragePlugin - gentics/mesh-incubator#111

## TODO

* [ ] Provide default (build-in) local storage support
      Move RangeRequestHandler code and refactor BinaryFieldResponseHandler
* [ ] Refactor Range Request Handling (to use binary storage)
      Code should be part of the local storage provider implementation
* [ ] Merge https://github.com/gentics/mesh/pull/868 with PR
* [ ] Merge parts of https://github.com/gentics/mesh/compare/dev-new-es-handling-upload-s3...jdbranham:dev-new-es-handling-upload-s3
* [ ] Add BinaryProcessorPlugin support / tests
* [ ] Add BinaryStoragePlugin support / tests
* [ ] Add BinaryAuxSourcePlugin support / tests
* [ ] Create S3 storage plugin
      Move S3 provider code into plugin
* [ ] Documentation
      Document new plugin types
* [ ] Plugin Examples
      Virus Scanner plugin
      External binary source plugin

## Limitations

* Migration between binary storage systems will not be supported out of the box

## Questions

* Should webroot requests be answered via proxying the request between client and S3 (Additional traffic) or redirect via `302` to S3 URL.
* Should the webroot link resolver also render S3 urls whenever possible? This way a client request would directly hit S3 without involving the Mesh server.

### DIFF
--- a/common/src/main/java/com/gentics/mesh/plugin/registry/binary/BinaryStoragePluginRegistry.java
+++ b/common/src/main/java/com/gentics/mesh/plugin/registry/binary/BinaryStoragePluginRegistry.java
@@ -1,0 +1,7 @@
+package com.gentics.mesh.plugin.registry.binary;
+
+import com.gentics.mesh.plugin.registry.PluginRegistry;
+
+public interface BinaryStoragePluginRegistry extends PluginRegistry {
+
+}
--- a/common/src/main/java/com/gentics/mesh/storage/BinaryStorage.java
+++ b/common/src/main/java/com/gentics/mesh/storage/BinaryStorage.java
@@ -30,23 +30,25 @@ public interface BinaryStorage {
 	 * @param stream
 	 * @param uuid
 	 *            Uuid of the binary to be stored
+	 * @param size
 	 * @param temporaryId
 	 * @return
 	 */
-	Completable storeInTemp(Flowable<Buffer> stream, String temporaryId);
+	Completable storeInTemp(Flowable<Buffer> stream, long size, String temporaryId);
 
 	/**
 	 * Store the stream directly.
 	 * 
 	 * @param stream
+	 * @param size
 	 * @param uuid
 	 * @return
-	 * @deprecated Use {@link #storeInTemp(Flowable, String, String)} in combination with {@link #moveInPlace(String, String)} instead.
+	 * @deprecated Use {@link #storeInTemp(Flowable, long, String)} in combination with {@link #moveInPlace(String, String)} instead.
 	 */
 	@Deprecated
-	default Completable store(Flowable<Buffer> stream, String uuid) {
+	default Completable store(Flowable<Buffer> stream, long size, String uuid) {
 		String id = UUIDUtil.randomUUID();
-		return storeInTemp(stream, id).andThen(moveInPlace(uuid, id));
+		return storeInTemp(stream, size, id).andThen(moveInPlace(uuid, id));
 	}
 
 	/**
--- a/common/src/main/java/com/gentics/mesh/storage/BinaryStorage.java
+++ b/common/src/main/java/com/gentics/mesh/storage/BinaryStorage.java
@@ -8,6 +8,7 @@ import com.gentics.mesh.util.UUIDUtil;
 
 import io.reactivex.Completable;
 import io.reactivex.Flowable;
+import io.reactivex.Single;
 import io.vertx.core.buffer.Buffer;
 
 /**
@@ -66,7 +67,7 @@ public interface BinaryStorage {
 	 * @param binaryField
 	 * @return
 	 */
-	boolean exists(BinaryGraphField field);
+	Single<Boolean> exists(BinaryGraphField field);
 
 	/**
 	 * Read the binary data which is identified by the given binary uuid.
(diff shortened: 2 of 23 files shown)


Jotschi:
## TODO:

* [ ] Config handling
* [ ] Selection of storage driver
* [ ] Refactor Range Request Handling (to use binary storage)
* [ ] Documentation
* [ ] Examples

jdbranham:
Hey @Jotschi - Is this something we could help with?
We're looking to deploy Mesh with AWS EKS, but EKS doesn't support EFS, so S3 would be perfect.

Jotschi:
@jdbranham Yes, any help would allow this feature to be added to mesh sooner.

I have just rebased the PR on the dev branch to get it up-to-date. We have however not yet decided how we want to add support to S3 to Mesh. In this PR I tried to add a new implementation of an `BinaryStorage` which is the implementation which takes care of handling store / retrieve operation for binaries.

Another option that is on the table is to extend the plugin support and provide ways to add binary handling for plugins. The idea is that this would allow a plugin to be written which can provide binary data by hooking into the binary delivery and store processes.

Completely independent from those decisions is however the implementation for the S3 connection.

It would help if the S3BinaryStorage implementation would be finished. If this is done in a working and resilient way it would take one task of the Todo list from this feature.

I think the `S3BinaryStorageTest` is a good way to start. It uses minio which is a open source S3 server. The test starts that server and tests the S3BinaryStorage implementation.

The whole implementation and test is very much Work In Progress. I wanted to get a better understanding how I could get all the needed parts aligned before starting work on the feature.

Resilience is an important part which has not yet been addressed by my WIP. In some S3 systems a HEAD request might not directly return a result when the object has just been created before. Retry handling for store and retrieve should also be added.

The API should be non-blocking and use `Flowable` for data handling where possible.

Gentics Mesh supports range requests. Thus it must also be possible to retrieve parts of a file. This aspect has not yet been added to the `BinaryStorage`.

Let me know if you have any question.

jdbranham:
Thanks, I'll definitely check it out and make an attempt.

Also, I just ran across this article. I'm not abandoning S3, but it's interesting...
https://aws.amazon.com/about-aws/whats-new/2019/09/amazon-eks-announces-beta-release-of-amazon-efs-csi-driver/

jdbranham:
I've got some changes, but I'm not sure if they are valid.  
I had some issues trying to pull the minio version from docker, but it seemed to work ok using the latest.  
Then I ran into an issues where the `head` requests return a `404` instead of the expected exceptions, so I modified the conditional statement to work for both scenarios.  
The tests pass, so I'm trying to figure out where to go next.  
Should I write a test that fails on the first attempt to create an object, but passes on a 'retry'?  
[I'll need to research how to simulate that one...]

Also, you mentioned 'range requests'. Is that something that will be added to the BinaryStorage interface?  
https://github.com/gentics/mesh/blob/dev/common/src/main/java/com/gentics/mesh/storage/BinaryStorage.java

Here's the comparison of my changes - 
https://github.com/gentics/mesh/compare/dev-new-es-handling-upload-s3...jdbranham:dev-new-es-handling-upload-s3

Jotschi:
@jdbranham I checked the diff. Looks good this far. I would however not use the docker latest tag anywhere. This can cause instability once the version is updated. Best to use a fixed tag.

Yes, the range request handling would need to be added to the binary storage. This might not be that easy since the LocalFilesystem Storage (current implementation) uses the NIO file send method. I guess it would be best to just provide the offset, source etc to the binary storage and the storage implementation can fulfill the request. 

The `RangeRequestHandlerImpl` is currently taking care of this task. It is however just taking the local file path from the `BinaryFieldResponseHandler#109`. Currently the `BinaryStorage#getLocalPath` method is used to get the local path. For S3 this is not possible.
I'm not sure whether the request should be proxied directly to S3 or whether a local disk-based cache should be involved. I think it might be best to keep the cache in mind but add it at a later date.

Regarding how to move on:

You could try to add toxiproxy to minio and check whether the implementation still works even when the connection is slow/unreliable.
See https://www.testcontainers.org/modules/toxiproxy/

We use this in `MeshTestContext#482` for example. In your case the minio container and the toxi container must be in the same network and you would not use the minio IP/Port but the toxi container proxy IP/Port to hook it to S3. The proxy object provides you with methods to taint the connection. 

Let me know if you have any other question.

ReneVolution:
Hey @Jotschi, has there been any progress on this since your last comment? I would be super keen for native (or plugin based) S3 support to appear in Mesh and wonder how much work is still left to be done. 

Thanks, René

Jotschi:
@ReneVolution I would also really like to have S3 support but things have been a little crazy lately. I'm not sure how much is actually left. I suspect this could be added in about a week (5 work days).

ReneVolution:
@Jotschi Absolutely understood. Thanks for the quick reply.

Jotschi:
Added plugin drafts for https://github.com/gentics/mesh-incubator/issues/205  https://github.com/gentics/mesh-incubator/issues/111 https://github.com/gentics/mesh-incubator/issues/27

Jotschi:
Branch should/could be combined with https://github.com/gentics/mesh/pull/868

PR #666, opened 2019-03-08
Dev user roles

### LINKED ISSUE #665: Add unique roles to User response
To facilitate user specific caching in the [Gentics Portal|Java](https://github.com/gentics/mesh-portal-framework), all active roles of a user should be available in a User response.

yrucrem:
Fixes #665

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
@@ -22,6 +22,7 @@ import static org.apache.commons.lang3.StringUtils.isEmpty;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.BooleanUtils;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
```diff
@@ -223,6 +224,12 @@ public User setEmailAddress(String emailAddress) {
 		return new TraversalResult<>(out(ASSIGNED_TO_ROLE).frameExplicit(RoleImpl.class));
 	}
 
+	@Override
+	public Page<? extends Role> getRolesViaShortcut(User user, PagingParameters params) {
+		VertexTraversal<?, ?, ?> traversal = out(ASSIGNED_TO_ROLE);
```
Jotschi: Check whether it is possible to use the index instead. This would speedup the GraphQL field retrieval for `.roles`.

Jotschi:
```
// GraphQL:
String indexName = "e." + ASSIGNED_TO_ROLE + "_out";
// userId: this.id()
new DynamicTransformablePageImpl<>(ac.getUser(), indexName, userId, RoleImpl.class, pagingInfo, perm, null, true);
```

Jotschi:
Add comment to UserImpl#getRolesViaShortcut: TODO Use shortcut index

core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
yrucrem: Done

PR #667, opened 2019-03-12
Add user roles hash to user responses

yrucrem:
* Remove roles from REST user response
* Add `rolesHash` to REST user response and GraphQL user fields

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
@@ -23,7 +23,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.Spliterator;
-import java.util.stream.Collector;
+import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 import org.apache.commons.lang3.BooleanUtils;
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/gentics/mesh/core/data/impl/UserImpl.java
```diff
@@ -214,6 +216,16 @@ public User setEmailAddress(String emailAddress) {
 		return new TraversalResult<>(out(HAS_USER).frameExplicit(GroupImpl.class));
 	}
 
+	@Override
+	public String getRolesHash() {
+		String indexName = "e." + ASSIGNED_TO_ROLE + "_out";
+		Spliterator<Edge> itemEdges = getGraph().getEdges(indexName.toLowerCase(), id()).spliterator();
+
+		return StreamSupport.stream(itemEdges, false)
+			.map(itemEdge -> itemEdge.getVertex(Direction.IN).getId().toString())
```
Jotschi: TODO:
* Sort Ids
* Concat Ids using joining
* Finally generate hash

rest-model/src/main/java/com/gentics/mesh/core/rest/user/UserResponse.java
```diff
@@ -40,14 +39,14 @@
 	@JsonPropertyDescription("Flag which indicates whether the user is enabled or disabled. Disabled users can no longer log into Gentics Mesh. Deleting a user user will not remove it. Instead the user will just be disabled.")
 	private Boolean enabled;
 
+	@JsonProperty(required = true)
+	@JsonPropertyDescription("Hash of roles the user has.")
```
Jotschi:
```suggestion
	@JsonPropertyDescription("Hashsum of user roles which can be used for user permission caching.")
```

PR #678, opened 2019-03-21
Add sonarqube to build pipeline

Jotschi:


Jotschi:
* Add dedicated execution for high memory tests.
* Remove dead comment from pom.xml
* Remove find sh in jenkins file

PR #691, opened 2019-04-02
Add event handling for node reference onDelete actions

Jotschi:


core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -1385,11 +1395,67 @@ public void delete(BulkActionContext bac, boolean ignoreChecks, boolean recursiv
...
+
+		// Check whether NodeGraphFields (in either NGFC, or Micronode) reference this node.
+		for (NodeGraphField field : inE(HAS_FIELD).has(NodeGraphFieldImpl.class).frameExplicit(NodeGraphFieldImpl.class)) {
+			for (GraphFieldContainer container : field.outV().frame(GraphFieldContainer.class)) {
+				followContainer(bac, container, handledNodeUuids);
+			}
+		}
+
+		// Check whether Node lists (in either NGFC, or Micronode) reference this node.
+		for (NodeGraphField listItem : inE(HAS_ITEM).has(NodeGraphFieldImpl.class).frameExplicit(NodeGraphFieldImpl.class)) {
```
philippguertler: We don't need to inspect the edge, use `in` instead

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -1385,11 +1395,67 @@ public void delete(BulkActionContext bac, boolean ignoreChecks, boolean recursiv
...
+		bac.add(onDeleted(getUuid(), getSchemaContainer(), null, null, null));
 		getElement().remove();
 		bac.process();
 	}
 
+	private void addReferenceUpdates(BulkActionContext bac) {
+		Set<String> handledNodeUuids = new HashSet<>();
+
+		// Check whether NodeGraphFields (in either NGFC, or Micronode) reference this node.
+		for (NodeGraphField field : inE(HAS_FIELD).has(NodeGraphFieldImpl.class).frameExplicit(NodeGraphFieldImpl.class)) {
```
philippguertler: We don't need to inspect the edge, use `in` instead

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -1385,11 +1395,67 @@ public void delete(BulkActionContext bac, boolean ignoreChecks, boolean recursiv
...
+			}
+		}
+	}
+
+	private void checkAndAddReferenceEvent(BulkActionContext bac, NodeGraphFieldContainer nodeContainer, Set<String> handledNodeUuids) {
+		for (GraphFieldContainerEdgeImpl edge : nodeContainer.inE(HAS_FIELD_CONTAINER).frameExplicit(GraphFieldContainerEdgeImpl.class)) {
+			ContainerType type = edge.getType();
+			// Only handle published or draft contents
+			if (type.equals(DRAFT) || type.equals(PUBLISHED)) {
+				if (nodeContainer.isDraft() || nodeContainer.isPublished()) {
```
philippguertler: remove double check

core/src/test/java/com/gentics/mesh/core/field/micronode/MicronodeFieldEndpointTest.java
```diff
@@ -212,6 +227,125 @@ public void testCreateNodeWithField() {
...
+		call(() -> client().updateMicroschema(vcardUuid, request));
+
+		// 1. Set the reference
+		MicronodeResponse field = new MicronodeResponse();
+		field.setMicroschema(new MicroschemaReferenceImpl().setName("vcard"));
+		field.getFields().put("firstName", new StringFieldImpl().setString("Max"));
+		field.getFields().put("lastName", new StringFieldImpl().setString("Moritz"));
+
+		NodeFieldListImpl nodeList = new NodeFieldListImpl();
+		nodeList.add(new NodeFieldListItemImpl().setUuid(targetUuid));
```
philippguertler: add additional entries

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
Jotschi: Not possible since we need to do a type check on the edge

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
Jotschi: Not possible since we need to do a type check on the edge

PR #694, opened 2019-04-03
Fix graphql tagged nodes perm

Jotschi:


common/src/main/java/com/gentics/mesh/core/data/Tag.java
```diff
@@ -111,4 +114,6 @@ default TypeInfo getTypeInfo() {
 	 */
 	Project getProject();
 
+	Stream<? extends Node> findTaggedNodesStream(InternalActionContext ac, GraphPermission permission);
```
philippguertler: Use TraversalResult

core/src/main/java/com/gentics/mesh/core/data/impl/GraphFieldContainerEdgeImpl.java
```diff
@@ -173,4 +176,20 @@ public LanguageRestrictionFunction(String languageTag) {
 		}
 	}
 
+	/**
+	 * Check whether the node has a field for the branch and given type.
```
philippguertler: Rewrite javadoc

Jotschi:
TODO: Refactor method and move main stream code into NodeImpl

PR #709, opened 2019-04-19
Make container flag in schema optional - Fixes #165

### LINKED ISSUE #165: Change schema update request defaults

Jotschi:


Jotschi:
buildnow

core/src/main/java/com/gentics/mesh/core/data/root/impl/SchemaContainerRootImpl.java
```diff
@@ -141,6 +141,11 @@ public SchemaContainer create(InternalActionContext ac, SearchQueueBatch batch,
 		MeshAuthUser requestUser = ac.getUser();
 		SchemaModel requestModel = JsonUtil.readValue(ac.getBodyAsString(), SchemaModelImpl.class);
 		requestModel.validate();
+
```
Jotschi: Don't throw an error. Instead use default `false` if the container flag is omitted.

core/src/test/java/com/gentics/mesh/core/schema/SchemaEndpointTest.java
```diff
@@ -96,6 +95,13 @@ public void testCreate() throws GenericRestException, Exception {
 		}
 	}
 
+	@Test
+	public void testCreateWithoutContainerFlag() {
+		SchemaCreateRequest createRequest = FieldUtil.createMinimalValidSchemaCreateRequest();
+		createRequest.setContainer(null);
+		call(() -> client().createSchema(createRequest), BAD_REQUEST, "schema_error_container_flag_missing");
```
Jotschi: Assert that  schema gets created with false flag.

rest-model/src/main/java/com/gentics/mesh/core/rest/schema/impl/SchemaCreateRequest.java
```diff
@@ -26,7 +26,7 @@
 
 	@JsonProperty(required = false)
 	@JsonPropertyDescription("Flag which indicates whether nodes which use this schema store additional child nodes.")
-	private boolean container = false;
+	private Boolean container = false;
```
Jotschi: Remove default value and set the default value in the schema create endpoint handler.

PR #712, opened 2019-04-23
Changes to GraphQL API for /api/v2

### LINKED ISSUE #317: Micronode GraphQL-Response uses different format
- v0.16.x

## Problem

Currently the issue lies in the Micronode handling in the GraphQL endpoint. The format is not the same as everywhere else, which causes confusion and additional normalization work (So you can actually compare the response values).

Normal Response (`POST search/nodes`, `GET nodes/<uuid>`, etc.): 
```json
{
    "uuid": "...",
    "microschema": {
        "name": "..."
    },
    "fields": {
        "myField": "..."
    }
}
```

GraphQL-Response:
```json
{
    "myField": "..."
}
```

GraphQL should return Micronodes the same way as every other Response which contains Micronodes

### LINKED ISSUE #428: GraphQL: Pull up schema types
# Description
Currently, the type for different schemas in GraphQL is differentiated in fields. So to query fields of a node, the query has to look like this:
```graphql
{
  node(path: "/images") {
    uuid
    fields {
      ... on folder {
        name
      }
    }
  }
}
```

I recommend changing the GraphQL types, so that the query looks like this instead:
```graphql
{
  node(path: "/images") {
    uuid
    ... on folder {
      fields {
          name
      }
    }
  }
}
```

The change was that the `... on folder` was pulled up one level.

A common use case is to have different queries for different schemas. With this change, it would be very easy to create a fragment for each schema and use them in the node query. Here is an example:

```graphql
{
  node(path: "/images") {
    uuid
    ... folder
    ... vehicleImage
  }
}

fragment folder on folder {
  fields {
    name
  }
}

fragment vehicleImage on vehicleImage {
  path
  fields {
    name
  }
}
```

As you can see, this leads to a nice and clean main query, and a fragment for each schema. Also this allows for schema fragments to query properties that are not fields, which is currently not possible.

# Backend Implementation
We could have a common interface for Nodes which includes all common fields (uuid, path, creator, etc.)
For each schema we then create an implementation of that interface with the additional `fields` property.

# API Implementation
Simply changing this would cause a breaking change. I see these ways to solve this:
1. Just let it be a breaking change. This is maybe acceptable for Mesh 1.0
2. Instead of breaking the existing API, add new types for the API (ex. NodeInterface, FolderNode, VehicleNode, etc.) and allow them to be queried in addition to the old types. Maybe we can deprecate the old API and later remove it.
3. Make the new API available through another configuration (mesh.yml, GET-Parameter, different URL)

philippguertler:
Resolves #317

Resolves #428


philippguertler:
Includes changes from #711

philippguertler:
TODO: grep repo for `/api/v1` or just `v1`

philippguertler:
buildnow

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/QueryTypeProvider.java
```diff
@@ -570,7 +566,10 @@ public GraphQLSchema getRootSchema(GraphQLContext context) {
 
 		additionalTypes.add(createLinkEnumType());
 
-		additionalTypes.addAll(nodeFieldTypeProvider.generateSchemaFieldTypes(context));
+		if (context.getApiVersion() >= 2) {
```
philippguertler: Use Versioned class to execute code conditionally

Jotschi:
buildnow

Jotschi:
buildnow

PR #719, opened 2019-04-26
Fix eventbus auto reconnect test

philippguertler:


Jotschi:
Remove `BootstrapInitializer#initProjects` and add code to `AdminHandler`

Issue #702, opened 2019-04-18
Content-Disposition header for binary downloads


philippguertler:
Currently every binary download response has the `Content-Disposition` header set to attachment, forcing the browser to download the file when linked directly. Some times it makes more sense to let the file be viewed directly in the browser.

Here are some suggestions:
* Remove the header and use the download attribute of the `<a>` tag to make the browser download the link (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes). This does not work cross-origin.
* Make the header inline for some hard coded content types (what you suggested)
* Make the headers somehow configurable - for some projects it makes sense to download a pdf file, for others it makes more sense to have it inline.

philippguertler:
From @xxyy in Gitter:

> I've just seen that Moodle seems to do the download thing by appending a query parameter ?forcedownload=1, which might also be an option, depending on why Content-Disposition was originally added / how it is currently used

xxyy:
It seems that the header was initially added in [this commit](https://github.com/gentics/mesh/commit/c8596bad93a8962c918e4a174372156ec95d3ee0), which sadly doesn't provide context as to why it was. It's not mentioned in the changelog either.

As @Jotschi mentioned on Gitter, it would probably make sense to align the behaviour to common web servers. Looking at the Google results for e.g. Apache, the only things that pop up are how to [make](https://serverfault.com/questions/101948/how-to-send-content-disposition-headers-in-apache-for-files) [it](https://stackoverflow.com/questions/3977159/set-content-disposition-header-to-attachment-only-on-files-in-a-certain-director) [add](https://serverfault.com/questions/101948/how-to-send-content-disposition-headers-in-apache-for%20-files) the header. Also, in the current version of the (main) source, the [only mentions of the header](https://github.com/apache/httpd/search?q=content-disposition&unscoped_q=content-disposition) are for form data. Similar results for [nginx](https://serverfault.com/questions/479468/content-disposition-header-on-nginx) and [IIS](https://blog.httpwatch.com/2010/03/24/four-tips-for-setting-up-http-file-downloads/). Hence, it seems like major web servers don't handle/send the header at all by default.

With this, not sending it might be a sensible default behaviour, but it really depends on which use-case caused it to be added in the first place.

Since, for the binary download request, there are already some query parameters, and controlling responses via query parameters seems to match the general spirit of the REST API, adding the query parameter switch seems like the best option to me, given what I know about the context, and also respecting that some people might already be depending on the existing behaviour.

[...]

PR #730, opened 2019-05-08
Add encoded filename in content-disposition header for BinaryFieldRes…

Alykoff:
This pull-request was inpired by 
https://github.com/gentics/mesh/issues/702

When we use other language than English lang for file names, we will be faced with a problem when we try to download file from mesh.

Снимок экрана 2019-05-08 в 17 31 26

After this pull request the download of files will be looks like that:

Снимок экрана 2019-05-08 в 17 50 56

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryFieldResponseHandler.java
+++ b/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryFieldResponseHandler.java
@@ -7,6 +7,7 @@ import static io.netty.handler.codec.http.HttpResponseStatus.NOT_MODIFIED;
 
 import javax.inject.Inject;
 import javax.inject.Singleton;
+import javax.mail.internet.MimeUtility;
 
 import com.gentics.mesh.context.InternalActionContext;
 import com.gentics.mesh.context.impl.InternalRoutingActionContextImpl;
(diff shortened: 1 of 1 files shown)


Jotschi:
Thanks for your PR. 

I checked my initial idea of removing the header all together but this would negatively affect the endpoint `GET /{project}/nodes/{nodeUuid}/binary/{fieldName}` (downloaed files would always be named by the fieldname)

The [RFC6266](https://tools.ietf.org/html/rfc6266) indicates that modern clients support RFC5987 encoded filenames. (See: https://tools.ietf.org/html/rfc6266#section-4.3)

I suggest to add the mentioned `filename*` parameter. See also https://stackoverflow.com/a/6745788/644326

Before I can merge the PR I have to run the tests and a Contributor License Agreement has to be accepted. We'll update you about the details since we are currently reworking the CLA process.

Alykoff:
@Jotschi thanks for your comment,

I have added support `filename*` parameter in new commit in this pull request.

mephinet:
@Alykoff Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our Contributor License Agreement before we can accept your contribution: https://cla-assistant.io/gentics/mesh?pullRequest=730

Jotschi:
@Alykoff I have rebased the PR and added some tests and merged it via https://github.com/gentics/mesh/pull/755. The fix will be added to the next release. Thanks

PR #739, opened 2019-05-14
Speedup attachment ingest handling

Jotschi:


GenticsDev:
Can't set status; build failed.

PR #757, opened 2019-06-04
Image manipulation returns same image with changed parameters

### LINKED ISSUE #669: Image manipulation returns same with changed parameters
- v0.30.2

## Problem
When changing the parameters of an image manipulation request, it can happen that an old cached image is returned even though the parameters have changed.

## Reproducer
1. Upload an image
2. Fetch the image like this: `http://localhost:8081/api/v1/test/nodes/fe7a22767bc14db7ba22767bc14db721/binary/binary?w=300&h=400&fpx=0.46&fpy=0.35&fpdebug=true&crop=fp&fpz=2`
3. Fetch it again and remove the fpz parameter: `http://localhost:8081/api/v1/test/nodes/fe7a22767bc14db7ba22767bc14db721/binary/binary?w=300&h=400&fpx=0.46&fpy=0.35&fpdebug=true&crop=fp`

I expected the zoomed out image to be returned, but instead I get the previous image again.

philippguertler:
Fixes #669

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -24,6 +24,11 @@ include::content/docs/variables.adoc-include[]
 * The support for the embedded Elasticsearch will be dropped in the future. It is highly recommended to link:{{< relref 
 "elasticsearch.asciidoc" >}}#_dedicated_elasticsearch[setup Elasticsearch as a dedicated service].
 
+[[v0.36.0]]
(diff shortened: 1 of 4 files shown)


Jotschi:
buildnow

Jotschi:
buildnow

Jotschi:
buildnow

PR #780, opened 2019-06-26
MADL - Stage 3

Jotschi:


core/src/main/java/com/gentics/mesh/core/data/root/impl/BranchRootImpl.java
```diff
@@ -53,7 +53,7 @@ public static void init(TypeHandler type, IndexHandler index) {
 
 	@Override
 	public Project getProject() {
-		return in(HAS_BRANCH_ROOT).nextOrDefaultExplicit(ProjectImpl.class, null);
+		return in(HAS_BRANCH_ROOT, ProjectImpl.class).nextOrNull();
```
philippguertler: Use next instead

madl/api/src/main/java/com/gentics/mesh/madl/traversal/TraversalResult.java
```diff
@@ -65,4 +65,13 @@ public boolean hasNext() {
 	public T next() {
 		return iterator().next();
 	}
+
+	public T nextOrNull() {
```
philippguertler: Consider this

core/src/main/java/com/gentics/mesh/core/data/node/impl/MicronodeImpl.java
```diff
@@ -101,8 +101,7 @@ public MicronodeResponse transformToRestSync(InternalActionContext ac, int level
 
 	@Override
 	public MicroschemaContainerVersion getSchemaContainerVersion() {
-		return out(HAS_MICROSCHEMA_CONTAINER).has(MicroschemaContainerVersionImpl.class).nextOrDefaultExplicit(MicroschemaContainerVersionImpl.class,
-				null);
+		return out(HAS_MICROSCHEMA_CONTAINER, MicroschemaContainerVersionImpl.class).nextOrNull();
```
philippguertler: Use next instead

PR #789, opened 2019-07-03
Fix parameters in mocked contexts

yrucrem:
The parameters map is always empty and not generated from the `query` parameter.


Jotschi:
@yrucrem will you are some more changes or can those be tested & merged?

PR #795, opened 2019-07-06
use junit test Rule for setting env variables

### LINKED ISSUE #794: OptionsLoaderTest.testApplyEnvs fails
- v0.36.5-SNAPSHOT [dev branch]

## Operating System
- Windows

## JVM
- Open JDK 11.0.3.7

## Problem
org.gentics.mesh.OptionsLoaderTest.testApplyEnvs fails 

## Reproducer
Fresh checkout of dev branch -> run JUnit tests on org.gentics.mesh.OptionsLoaderTest

Expected OptionsLoader to use the env variables.  
Actual behavior - the environment variables are not set, the OptionsLoader doesn't use them.

I've fixed this in a branch that uses a contributed JUnit package - 
https://stefanbirkner.github.io/system-rules/#EnvironmentVariables

The dependency looks well maintained, and is able to revert changes after the test, since it uses a JUnit Rule.

I can create a PR with this fix [added dependency], or we can try to understand why the reflective 'set' method in the class is failing and fix it.

jdbranham:
Fixes #794

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -181,7 +181,6 @@
 		<dependency>
 			<groupId>com.github.stefanbirkner</groupId>
 			<artifactId>system-rules</artifactId>
-			<version>1.19.0</version>
 			<scope>test</scope>
 		</dependency>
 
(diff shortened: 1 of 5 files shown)


Jotschi:
buildnow

Jotschi:
Thanks for the PR.

PR #796, opened 2019-07-06
Dev issue 715

### LINKED ISSUE #715: Plugin Admin Client uses wrong host
The creation of the admin client will correctly not work when the mesh configuration uses host 0.0.0.0 - This host can't be used to connect to. It is only used to bind to.

https://github.com/gentics/mesh/blob/b9e0f91a2091b958a2e255df45cf92a2346c46c3/plugin-api/src/main/java/com/gentics/mesh/plugin/AbstractPluginVerticle.java#L170

jdbranham:
The admin client will use 127.0.0.1 if the server was bound to 0.0.0.0

fixes #715

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -177,6 +177,13 @@
 			<type>test-jar</type>
 			<scope>test</scope>
 		</dependency>
+		
+		<dependency>
+			<groupId>com.github.stefanbirkner</groupId>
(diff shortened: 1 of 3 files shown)


Jotschi:
buildnow

Jotschi:
Merged and rebased via #804 - Thanks for the PR.

PR #799, opened 2019-07-08
Add test for recursively deleting a node in a branch

philippguertler:


core/src/test/java/com/gentics/mesh/core/node/NodeDeleteBranchEndpointTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+	@Test
+	public void deleteNodeInBranch() {
+		grantAdminRole();
+		NodeResponse parent = createNode();
+		NodeResponse otherParent = createNode();
+		publishNode(parent);
+		publishNode(otherParent);
+
+		waitForJobs(() -> {
```
Jotschi: Use waitFor instead.

PR #814, opened 2019-07-12
Add equals, hashCode to REST Model classes

Jotschi:


Jotschi:
buildnow

Jotschi:
* Add dedicated equals implementation for node response (use version, lang, uuid for comparison)

Jotschi:
buildnow

PR #820, opened 2019-07-18
Improve element load performance by avoiding OrientDB record loading

Jotschi:


core/src/main/java/com/gentics/mesh/core/data/root/impl/ProjectRootImpl.java
```diff
@@ -47,6 +50,19 @@
  */
 public class ProjectRootImpl extends AbstractRootVertex<Project> implements ProjectRoot {
 
+	public static final EventAwareCache<String, Project> PROJECT_CACHE = EventAwareCache.<String, Project>builder()
+		.events(PROJECT_DELETED, PROJECT_UPDATED)
+		.action((event, cache) -> {
+			System.out.println(event.body().encodePrettily());
```
Jotschi: Remove this!

Jotschi: Name this PROJECT_NAME_CACHE

ferma/src/main/java/com/syncleus/ferma/AbstractEdgeFrame.java
```diff
@@ -38,7 +41,11 @@
 
 	@Override
 	public Edge getElement() {
-		return (Edge) super.getElement();
+		Element e = super.getElement();
```
Jotschi: Use the old implementation here

ferma/src/main/java/com/syncleus/ferma/AbstractVertexFrame.java
```diff
@@ -40,11 +42,15 @@
 
 	@Override
 	public Vertex getElement() {
-		if (super.getElement() instanceof WrappedElement) {
-			return (Vertex) ((WrappedElement) super.getElement()).getBaseElement();
-		} else {
-			return (Vertex) super.getElement();
+		Element e = super.getElement();
```
Jotschi: Remove null check

PR #821, opened 2019-07-18
Fix newPassword handling

Jotschi:


Jotschi:
We instead added an error to the login process that informs about the conflicting parameters.

PR #855, opened 2019-08-20
Add generic OAuth2/OIDC support

Jotschi:
Add generic support for OAuth2 to also support other providers and not only keycloak


Jotschi:
TODO:
* [ ] Changelog
* [ ] Update docs
* [ ] Update example plugins
* [ ] Check other oauth2 tests
* [ ] Test Auth0

PR #870, opened 2019-09-02
Fix permissions and add RolePerms to GraphQL

philippguertler:


common/src/main/java/com/gentics/mesh/core/data/MeshCoreVertex.java
```diff
@@ -50,12 +51,11 @@
 	void fillCommonRestFields(InternalActionContext ac, FieldsSet fields, GenericRestResponse model);
 
 	/**
-	 * Add role permissions to given rest model object.
+	 * Get the permissions of a role for this element.
 	 * 
-	 * @param ac
-	 * @param model
+	 * @param roleUuid
```
Jotschi: Missing @return @param

philippguertler:
buildnow

PR #877, opened 2019-09-09
Change TX retry delay default to 10ms

yrucrem:


api/src/main/java/com/gentics/mesh/etc/config/GraphStorageOptions.java
```diff
@@ -53,13 +55,18 @@
 	@EnvironmentVariable(name = MESH_GRAPH_SYNC_WRITES_ENV, description = "Override the graph database sync writes flag.")
 	private boolean synchronizeWrites = DEFAULT_SYNC_WRITES;
 
+	@JsonProperty(defaultValue = DEFAULT_TX_RETRY_DELAY + "ms")
+	@JsonPropertyDescription("The delay in milliseconds when a transaction has to be retried.")
+	@EnvironmentVariable(name = MESH_GRAPH_TX_RETRY_DELAY_ENV, description = "Override the transaction retry delay.")
```
Jotschi: Add default to description text

doc/src/main/docs/administration-guide.asciidoc
```diff
@@ -222,7 +222,7 @@ include::content/docs/snippets/config/storage.inc[]
 |======
 | Configuration                                   | Type    | Default        | Description
 | ```ridBag.embeddedToSbtreeBonsaiThreshold```    | Number  | 2147483647     | Configures the ridbag threshold for OrientDB. This setting controls how OrientDB manages the internal ridbag data structure. This setting will be ignored when run in clustered mode. See OrientDB documentation for more details.
-| ```tx.retry.delay```                            | Number  | 0              | Sets the initial transaction retry delay in milliseconds which is applied when retrying failed transactions due to concurrent changes. A value of 0 will disable the delay.
+| ```tx.retry.delay```                            | Number  | 10             | Sets the initial transaction retry delay in milliseconds which is applied when retrying failed transactions due to concurrent changes. A value of 0 will disable the delay.
```
Jotschi: Remove this.

api/src/main/java/com/gentics/mesh/etc/config/GraphStorageOptions.java
yrucrem: Done.

PR #883, opened 2019-09-24
fix: update field operation is now correctly migrating the value of t…

cschockaert:
…he field data

when updating a schema field data is loss:

step to reproduce

create a schema with a field named 'bad_name'
create a node with this schema add a value to the field bad_name

migrate the field with this operation:
```
	{
	  "changes": [
	    {
	      "operation": "UPDATEFIELD",
	      "properties": {
	        "field": "bad_name",
	        "name": "realname"
	      }
	    }
	  ]
	}
```

then migrate the branch schema to latest

the node is migrated to the last schema (update) BUT the field data of 'realname' is not populated by the old value of field "bad_name"


Jotschi:
Included in https://github.com/gentics/mesh/pull/884

Jotschi:
Merged via #884

PR #884, opened 2019-09-24
Fix field rename operation

Jotschi:


Jotschi:
TODO:
* Review actual change
* Check why the old string field is still present (may be due to NGFC copy operation which clones all properties)

PR #887, opened 2019-10-01
Make index sync at startup non blocking

### LINKED ISSUE #862: Don't block Mesh startup when index sync fails
In some cases a forced index sync is being invoked. This will cause the reindex action in BootstrapInitializerImpl. This action will block until all documents have been synced. This will stall the startup process if ES is broken / unreachable. We should not block the startup in this case.

philippguertler:
Fixes #862


core/src/main/java/com/gentics/mesh/cli/BootstrapInitializerImpl.java
```diff
@@ -185,9 +185,7 @@
 			return;
 		}
 		// Ensure indices are setup and sync the documents
-		log.info("Invoking index sync. This may take some time..");
```
Jotschi: Why did you remove this logging stuff?

PR #893, opened 2019-10-10
Reduce memory by removing nested graph element reference

Jotschi:


madl/api/src/main/java/com/syncleus/ferma/AbstractEdgeFrame.java
```diff
@@ -38,7 +39,7 @@
 
 	@Override
 	public Edge getElement() {
-		return (Edge) super.getElement();
+		return null;
```
philippguertler: Remove this method

madl/api/src/main/java/com/syncleus/ferma/AbstractVertexFrame.java
```diff
@@ -41,12 +39,7 @@
 
 	@Override
 	public Vertex getElement() {
-		Element e = super.getElement();
-		if (e instanceof WrappedElement) {
-			return (Vertex) ((WrappedElement) e).getBaseElement();
-		} else {
-			return (Vertex) e;
-		}
+		return null;
```
philippguertler: Remove this method

Issue #904, opened 2019-10-17
Schemafield label property not resettable


bernhardriegler:
## Gentics Mesh Version, operating system, or hardware.
1.0.2

## Operating System
-

## Problem
Regarding Mesh REST API endpoint: ```POST /schemas/schemaUuid```:
Once a label for a field is set, it can not be removed.

### Current Behavior (sequence):
1. Request payload schema.fields[0]: 
```
{
	  "name": "testField",
	  "label": null,
	  "type": "string"
  }
```
Result:
Schema still does contain the label property for this field

### Expected Behavior:
by posting the value null for a label property it should be removed from the schema

Jotschi:
Related to https://github.com/gentics/mesh/pull/905

philippguertler:
Is it difficult in Java to differentiate between unset properties and properties set to null.
When updating without setting the label at all, we expect no changes to the label. We treat null and unset as the same here, that's why it is not possible to remove the label.

If we were to differentiate here, we would have to do the same for many other places in Mesh. This would be a huge change, which is also breaking.

As an alternative, you can set the label to an empty string.

PR #905, opened 2019-10-17
Fix Schema Label Null Handling

Jotschi:


Jotschi:
Setting the property in the graph to null is not enough. It seems the code that applies the value also needs to be adapted. A test has been added to `SchemaChangesEndpointTest#testResetFieldLabel`

philippguertler:
See this comment for more details: https://github.com/gentics/mesh/issues/904#issuecomment-624558424

PR #910, opened 2019-10-21
Added basic API Version enum and Version select

SiNEXiST:
A new simple Enum which holds the basePaths for the two different API Versions.
Additionally added optional parameters to the PluginEvironment to specify the version of the API you want to use.

Addresses the mesh-incubator#216 issue


plugin-api/src/main/java/com/gentics/mesh/plugin/env/PluginEnvironment.java
```diff
@@ -25,12 +26,19 @@
...
 	/**
-	 * Create a new admin client.
+	 * Create a new admin client with api version 1.
 	 * 
 	 * @return
 	 */
 	MeshRestClient createAdminClient();
 
+	/**
+	 * Creates a new admin client from the specified version.
```
Jotschi:
```suggestion
	 * Creates a new admin client for the specified version.
```

plugin-api/src/main/java/com/gentics/mesh/plugin/env/PluginEnvironment.java
```diff
@@ -25,12 +26,19 @@
...
+	 * 
+	 * @return
```
Jotschi: Add param - add return doc. This is a public API method and thus should have full javadoc

rest-client/src/main/java/com/gentics/mesh/rest/RestAPIVersion.java
```diff
@@ -0,0 +1,16 @@
+package com.gentics.mesh.rest;
+
+public enum RestAPIVersion {
```
Jotschi: I think there is no reason to not move this to `api` maven module.

PR #911, opened 2019-10-21
Cleaned up Option Objects

### LINKED ISSUE #900: Add Getters/Setters for all Options
## Problem

Some properties of the [MeshOptions](https://github.com/gentics/mesh/blob/dev/api/src/main/java/com/gentics/mesh/etc/config/MeshOptions.java) don't have a getter/setter. This makes it annoying to work with when setting up mesh programmatically, as some options simply can't be set.

As a workaround, I currently include a copy of the MeshOptions class with the additional setters to my project (as it's only to start it locally).

A common property as example would be the `defaultLanguage` Property.

All properties of the MeshOptions - as well as all referenced Classes like `HttpServerConfig` - have getters/setters for all properties

SiNEXiST:
Resolves #900


Jotschi:
Please update the java doc to the methods that have java doc. The return fluent api is missing. This will possibly break the build.

Jotschi:
Check build error and merge if ok.

PR #915, opened 2019-10-30
Make Rest API verticle amount configurable

philippguertler:


api/src/main/java/com/gentics/mesh/etc/config/HttpServerConfig.java
```diff
@@ -36,6 +36,8 @@
 	public static final String MESH_HTTP_SSL_KEY_PATH_ENV = "MESH_HTTP_SSL_KEY_PATH";
 	public static final String MESH_HTTP_CORS_ALLOW_CREDENTIALS_ENV = "MESH_HTTP_CORS_ALLOW_CREDENTIALS";
 
+	public static final int DEFAULT_VERTICLE_AMOUNT = 16;
```
Jotschi: Better to use default based on core count.

core/src/main/java/com/gentics/mesh/cli/CoreVerticleLoader.java
```diff
@@ -81,91 +66,45 @@ public void loadVerticles(List<String> initialProjects) {
...
 
-	private ElasticsearchProcessVerticle resetSearchVerticle() {
+	private void deploySearchVerticle() {
 		// Only deploy search sync verticle if we actually have a configured ES
 		ElasticSearchOptions searchOptions = meshOptions.getSearchOptions();
 		if (searchOptions != null && searchOptions.getUrl() != null) {
-			elasticsearchProcessVerticle = elasticsearchProcessVerticleProvider.get();
-			return elasticsearchProcessVerticle;
+			rxVertx.rxDeployVerticle(elasticsearchProcessVerticle, new DeploymentOptions()
+				.setInstances(1))
```
Jotschi: Why only one instance?

philippguertler: Before that change, there was also only one instance deployed.

common/src/main/java/com/gentics/mesh/router/RouterStorage.java
```diff
@@ -43,6 +44,7 @@
  * Project routers are automatically bound to all projects. This way only a single node verticle is needed to handle all project requests.
  * 
  */
+@Singleton
```
Jotschi:
```suggestion
```
Be aware that adding `singleton` will affect how handlers will be used across verticles. In general it is bad practice to re-use handlers across verticles/threads. Most Vert.x handlers are not written with thread safety in mind.

PR #917, opened 2019-10-31
GraphQL wait query-parameter

### LINKED ISSUE #805: GraphQL Search - Add wait support
When invoking a search query the `?wait` parameter will not be handled. We maybe should instead add another argument to the nodes call to provide a way to add the wait call.

SiNEXiST:
Added a basic implementation to support the `wait` query-parameter from the search endpoint.
Implementation is request-wide, as single query level doesn't work with the current synchronous GraphQL implementation.

Resolves #805


core/src/test/java/com/gentics/mesh/core/graphql/GraphQLWaitSearchEndpointTest.java
```diff
@@ -0,0 +1,66 @@
...
+	public void setUp() {
+		createNode(FIELD_NAME, new StringFieldImpl().setString(FIELD_VALUE));
+	}
+
+	@Test
+	public void queryWithWait() {
+		GraphQLResponse response = call(() -> client().graphqlQuery(PROJECT_NAME, getGraphQLQuery(QUERY_NAME),
+				new SearchParametersImpl().setWait(true)));
+		JsonObject json = new JsonObject(response.toJson());
+		System.out.println(json.encodePrettily());
```
philippguertler:
```suggestion
```

core/src/test/java/com/gentics/mesh/core/graphql/GraphQLWaitSearchEndpointTest.java
```diff
@@ -0,0 +1,66 @@
...
+		assertThat(elements.size()).isEqualTo(1);
+		assertThat(elements.getJsonObject(0).getJsonObject("fields").getString(FIELD_NAME)).isEqualTo(FIELD_VALUE);
+	}
+
+	@Test
+	public void queryWithoutWait() {
+		GraphQLResponse response = call(() -> client().graphqlQuery(PROJECT_NAME, getGraphQLQuery(QUERY_NAME),
+				new SearchParametersImpl().setWait(false)));
+		JsonObject json = new JsonObject(response.toJson());
+		System.out.println(json.encodePrettily());
```
philippguertler:
```suggestion
```

PR #918, opened 2019-11-04
Fix storing uploads in eventloop thread

philippguertler:


Jotschi:
Please check tests, add changelog and merge.

PR #920, opened 2019-11-05
Replace creator and editor edge with uuid

philippguertler:


core/src/main/java/com/gentics/mesh/core/data/generic/AbstractMeshCoreVertex.java
```diff
@@ -33,6 +34,9 @@
 import com.gentics.mesh.parameter.value.FieldsSet;
 import com.gentics.mesh.util.ETag;
 
+import io.netty.handler.codec.http.QueryStringDecoder;
```
Jotschi: Revert this

PR #926, opened 2019-11-09
Added an (simplified) Chinese i18n multilingual support.

Yuanuo:
Hello there,

First of all, thank you very much for sharing such a great project. Really very good.

Recently, I tried to add a Chinese (Simplified Chinese) version to the mesh and mesh-ui i18n. I also tried to share it. If anyone needs it, I hope this will bring you some help.

There are some instructions:
1, based on the git version of 2019-11-09;
2, each i18n property has been translated, no missing one;
3, I have personally checked each one, and probably there should be no problem other than the individual translation (such as "segment field").
4, of course, if you make some adjustments during the use process, there is no problem.


Jotschi:
buildnow

PR #932, opened 2019-11-15
Replace binary edges

philippguertler:


common/src/main/java/com/gentics/mesh/core/data/relationship/GraphRelationships.java
```diff
@@ -78,7 +78,6 @@ public static void init(TypeHandler type, IndexHandler index) {
 
 		// Binary
 		type.createType(edgeType(HAS_BINARY_ROOT));
```
Jotschi: Remove the edge and the edge type as well

common/src/main/java/com/gentics/mesh/core/data/relationship/GraphRelationships.java
```diff
@@ -148,7 +147,6 @@ public static void init(TypeHandler type, IndexHandler index) {
 
 	// Asset root
 	public static final String HAS_BINARY_ROOT = "HAS_BINARY_ROOT";
```
Jotschi: Remove this

core/src/main/java/com/gentics/mesh/core/data/binary/impl/BinariesImpl.java
```diff
@@ -0,0 +1,48 @@
...
+			binary.setSHA512Sum(sha512sum);
+			binary.setSize(size);
+			binary.setUuid(uuid);
+			return binary;
+		});
+	}
+
+	@Override
+	public Transactional<Binary> findByHash(String hash) {
+		return database.transactional(tx -> toStream(database.getVertices(BinaryImpl.class, new String[] { Binary.SHA512SUM_KEY }, new String[] { hash }))
```
Jotschi: I have added database.getVerticesTraversal - this might make this easier.

PR #945, opened 2019-11-25
Add health routes to rest API

philippguertler:


core/src/main/java/com/gentics/mesh/core/endpoint/admin/HealthEndpoint.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.gentics.mesh.auth.MeshAuthChain;
+import com.gentics.mesh.core.endpoint.handler.MonitoringCrudHandler;
+import com.gentics.mesh.rest.InternalEndpointRoute;
+import com.gentics.mesh.router.route.AbstractInternalEndpoint;
+
+public class HealthEndpoint extends AbstractInternalEndpoint {
+	private final MonitoringCrudHandler monitoringCrudHandler;
+
+	@Inject
```
Jotschi: Add constructor for RAML and add the endpoint to RAML as well.

PR #954, opened 2019-12-02
Update PF4J

Jotschi:


common/src/main/java/com/gentics/mesh/router/route/AbstractInternalEndpoint.java
```diff
@@ -49,7 +49,7 @@ public void init(Vertx vertx, RouterStorage rs) {
 	 */
 	protected void secureAll() {
 		if (chain != null) {
-			chain.secure(getRouter().route("/*"));
+			chain.secure(getRouter().route());
```
philippguertler: Does this have to do anything with the update?

PR #959, opened 2019-12-04
Add dedicated security logging

philippguertler:


common/src/main/java/com/gentics/mesh/router/route/FailureHandler.java
```diff
@@ -90,8 +91,8 @@ public void handle(RoutingContext rc) {
 			if (log.isDebugEnabled()) {
 				log.debug("Got failure with 401 code.");
 			}
-			InternalActionContext ac = new InternalRoutingActionContextImpl(rc);
 			String msg = I18NUtil.get(ac, "error_not_authorized");
+			ac.getSecurityLogger().info("Access to resource denied.");
```
Jotschi: I would prefer an API like:

```java
public static final SecurityLogger secLog = new SecurityLogger()
…
secLog.info(ac, "Access to resource denied")
```

PR #962, opened 2019-12-04
Remove embedded ES from mesh-server

Jotschi:


Jotschi:
@philippguertler PR was greenlit

PR #964, opened 2019-12-05
Cleanup Auth Code

philippguertler:


Jotschi:
If you want to really clean this up you should remove the auth chain and merge both handlers into one. Or at least avoid that the token gets decoded via two different JWT instances.

PR #972, opened 2019-12-11
Return 404 if UUID of another type is requested

philippguertler:


core/src/main/java/com/gentics/mesh/core/data/root/impl/NodeRootImpl.java
```diff
@@ -89,10 +89,14 @@ public String getRootLabel() {
 
 	@Override
 	public TraversalResult<? extends Node> findAll() {
-		ProjectImpl project = in(HAS_NODE_ROOT, ProjectImpl.class).next();
+		ProjectImpl project = getProject();
 		return project.findNodes();
 	}
 
+	private ProjectImpl getProject() {
```
Jotschi: Please only return interface type

core/src/main/java/com/gentics/mesh/core/data/root/impl/MeshRootImpl.java
```diff
@@ -377,7 +377,7 @@ public MeshVertex resolvePathToElement(String pathToElement) {
 
 	@Override
 	public Node findNodeByUuid(String uuid) {
-		return db().index().findByUuid(NodeImpl.class, uuid);
```
Jotschi: Check usage of `db().index().findByUuid`

CHANGELOG.adoc
```diff
@@ -14,6 +14,12 @@ include::content/docs/variables.adoc-include[]
 
 = Gentics Mesh Changelog
 
+[[Unreleased]]
+
+icon:check[] Fixed a bug in the `{projectX}/nodes/{nodeUUIDY}` that allowed nodes to be found even if `nodeUUIDY` is not part of `projectX`.
```
Jotschi: in the what?

PR #988, opened 2020-01-16
Enhance OpenShift Support

Jotschi:


Jotschi:
UID handling has been implemented according to https://docs.openshift.com/container-platform/3.9/creating_images/guidelines.html

server/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -o nounset
+set -o errexit
+
+UID=$(id -u)
```
philippguertler: This is the ID of the current user...

server/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+if ! whoami &> /dev/null; then
+  if [ -w /etc/passwd ]; then
+    echo "Creating mesh user uid to $UID"
+    echo "${USER_NAME:-mesh}:x:$UID:0:${USER_NAME:-mesh} user:/mesh:/sbin/nologin" >> /etc/passwd
```
philippguertler: ...which is then used here to create a new user with the same ID. Won't this result in two users with the same ID?

server/Dockerfile
```diff
@@ -1,53 +1,39 @@
...
-    && chown mesh: /elasticsearch -R
-
-USER mesh
+ADD ./target/mesh-server*jar /mesh/mesh.jar
+RUN mkdir /graphdb   && chown 1000:0 /graphdb  -R && chmod 770 /graphdb  && \
+    mkdir /uploads   && chown 1000:0 /uploads  -R && chmod 770 /uploads  && \
+    mkdir /backups   && chown 1000:0 /backups  -R && chmod 770 /backups  && \
+    mkdir /plugins   && chown 1000:0 /plugins  -R && chmod 770 /plugins  && \
+    mkdir /keystore  && chown 1000:0 /keystore -R && chmod 770 /keystore && \
+    mkdir /config    && chown 1000:0 /config   -R && chmod 770 /config && ln -s /config /mesh/config && \
```
philippguertler: Do the `chown` and `chmod` have any effect if the folders are overridden with mounted volumes? I thought the permissions from the host volume will be used in that case.

elasticsearch/src/main/java/com/gentics/mesh/search/ElasticsearchProcessManager.java
```diff
@@ -206,7 +206,8 @@ private String getJavaBinPath() throws FileNotFoundException {
 	private File prepareESDirectory() throws IOException, ZipException {
 		File outputDir = new File(".");
 		File esDir = new File(outputDir, OUTPUT_FOLDER_NAME);
-		if (!esDir.exists()) {
+		File esBin = new File(esDir, "bin");
+		if (!esBin.exists()) {
 			unzip("/" + RESOURCE_NAME + ".zip", outputDir.getAbsolutePath());
```
philippguertler: This can potentially take a long time now. Could this run in any timeouts on slow IO systems?

server/entrypoint.sh
Jotschi: The user with the given Id may not have a passwd entry. Thus whoami would fail. If whoami fails we need to create the entry in passwd to assign a "username" to this UID. Otherwise the process would just run with a user without a "real" name.

server/Dockerfile
Jotschi: No - This has only effect for non-mounted folders. When using OpenShift you usually run the pod with a random uid and mount/map config files (e.g. /config/mesh.yaml). The folder itself is however not mounted and thus writing to the folder would fail without chmod and chown.

elasticsearch/src/main/java/com/gentics/mesh/search/ElasticsearchProcessManager.java
Jotschi: I don't think this is relevant since we will drop ES embedded support in the future. The default mode for embedded ES is already set to false.

Jotschi:
TODO:
* Check whether setting USER=mesh env would be an alternative to adding the passwd entry

PR #992, opened 2020-01-18
Add checks and fallback handling DirectIO filesystem capabilities

Jotschi:


Jotschi:
TODO: skip checks when system property has already been set from outside

philippguertler:
Will this work on Mac/Windows?

PR #993, opened 2020-01-19
Add resolving for nodeName setting

Jotschi:
When deploying Mesh in K8S it may be desired to use the Pod Name as the node-name. This is especially handy when using stateful sets. Otherwise you would end up with a lot of node entries in OrientDB cluster configuration. Clustering would break when using writeQuorum majority and restarting Pods because new cluster nodes would be listed in the distributed configuration.

### DIFF
--- a/common/src/main/java/com/gentics/mesh/OptionsLoader.java
+++ b/common/src/main/java/com/gentics/mesh/OptionsLoader.java
@@ -23,6 +23,7 @@ import com.gentics.mesh.cli.MeshCLI;
 import com.gentics.mesh.cli.MeshNameProvider;
 import com.gentics.mesh.etc.config.MeshOptions;
 import com.gentics.mesh.util.UUIDUtil;
+import com.kstruct.gethostname4j.Hostname;
 
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
@@ -53,9 +54,17 @@ public final class OptionsLoader {
 		MeshOptions options = loadMeshOptions(defaultOption);
 		applyNonYamlProperties(defaultOption, options);
 		applyEnvironmentVariables(options);
+		applyReplacements(options);
 		applyCommandLineArgs(options, args);
 		options.validate();
 		return options;
+	}
+
+	private static void applyReplacements(MeshOptions options) {
+		String nodeName = options.getNodeName();
+		if ("$HOSTNAME".equals(nodeName)) {
+			options.setNodeName(Hostname.getHostname());
+		}
 	}
 
 	private static void applyNonYamlProperties(MeshOptions defaultOption, MeshOptions options) {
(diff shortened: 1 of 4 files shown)


Jotschi:
@philippguertler Maybe ```@HOSTNAME@``` or something similar would be better. Otherwise it would perhaps needed to be escaped in various places because of the dollar sign. WDYT?

common/src/test/java/com/gentics/mesh/OptionsLoaderTest.java
```diff
@@ -75,6 +76,16 @@ public void testApplyEnvs() throws Exception {
 		assertEquals(MappingMode.STRICT, options.getSearchOptions().getMappingMode());
 	}
 
+	@Test
+	public void testNodeNameWithHostnameVar() {
+		environmentVariables.set(MeshOptions.MESH_NODE_NAME_ENV, "$HOSTNAME");
+		MeshOptions options = OptionsLoader.createOrloadOptions();
+		String nodeName = options.getNodeName();
+		System.out.println(nodeName);
```
philippguertler:
```suggestion
```

common/src/main/java/com/gentics/mesh/OptionsLoader.java
```diff
@@ -53,11 +54,19 @@ public static MeshOptions createOrloadOptions(MeshOptions defaultOption, String.
...
 		applyEnvironmentVariables(options);
+		applyReplacements(options);
 		applyCommandLineArgs(options, args);
 		options.validate();
 		return options;
 	}
 
+	private static void applyReplacements(MeshOptions options) {
+		String nodeName = options.getNodeName();
+		if ("$HOSTNAME".equals(nodeName)) {
```
philippguertler: I think replacing the variable instead would be more intuitive. This is how it works everywhere else.

Jotschi:
> I agree, some other form of replacing the variable could be better. Have you researched how other projects do this?

I'm not sure how ES does it. They however use the metadata.name as node.name - See https://github.com/elastic/helm-charts/blob/d22b3379930752b7b215cf983414cf330eda143c/elasticsearch/templates/statefulset.yaml#L237

Maybe @berni2288 has an Idea how this is supposed to work?

Jotschi:
> hostname = podname
> 
> The podname of a statefulset is generated at runtime, so you can't inject it with Helm.
> However there is the env var HOSTNAME in the container, it could be used in the entrypoint script to configure the nodename. Or maybe the HOSTNAME env var can be directly read in the Mesh configuration?

@berni2288  Yes but why does ES use the`metadata.name` value of the pod? I don't understand how that could generate individual names for each pod.

Jotschi:
@berni2288 No. I was referring to way how ES handles the nodeName in their helm chart:

See: 
https://github.com/elastic/helm-charts/blob/d22b3379930752b7b215cf983414cf330eda143c/elasticsearch/templates/statefulset.yaml#L237

berni2288:
See: https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#capabilities-of-the-downward-api

Jotschi:
Usage of the downward api is a better solution.

PR #997, opened 2020-01-23
REST API Startup Enhancements

Jotschi:
# Speedup REST Startup

## Abstract

Reduces calls to generate JsonSchema and also uses static class to provide examples.

## Checklist

* [x] Added abstract that describes the change. 
* [x] Added changelog entry to `/CHANGELOG.adoc`.
* [x] Ensured that the change is covered by tests.

### DIFF
--- a/common/src/main/java/com/gentics/mesh/rest/impl/InternalEndpointRouteImpl.java
+++ b/common/src/main/java/com/gentics/mesh/rest/impl/InternalEndpointRouteImpl.java
@@ -10,7 +10,6 @@ import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/gentics/mesh/rest/impl/InternalEndpointRouteImpl.java
```diff
@@ -52,6 +53,8 @@
 
 	private static final Logger log = LoggerFactory.getLogger(InternalEndpointRoute.class);
 
+	private static final Map<Class<?>, String> SCHEMA_CACHE = Collections.synchronizedMap(new HashMap<>());
```
philippguertler: Use concurrent hashmap

common/src/main/java/com/gentics/mesh/rest/impl/InternalEndpointRouteImpl.java
```diff
@@ -330,6 +333,12 @@ public InternalEndpointRoute exampleResponse(HttpResponseStatus status, Object m
 		return this;
 	}
 
+	private String getSchema(Class<? extends Object> clazz) {
+		return SCHEMA_CACHE.computeIfAbsent(clazz, c -> {
```
philippguertler: Use method reference

PR #1012, opened 2020-02-20
Hotfix 1.4.x hz update

Jotschi:
## Abstract

Bump Hazelcast dependencies

### General

### On API Changes


Jotschi:
Plugin version is incompatible with 3.10.6. 

One change that causes the plugin to break is the refactoring of hazelcast:
https://github.com/hazelcast/hazelcast/commit/ae3919d27bdb3888f3a0265ca6c5023d7fe5a7ad#diff-eae50b0b5a0730d4b692e6f6cc53dba9

Earliest version that contains this change is 3.11.1

Jotschi:
We decided against this change in order to keep in-sync with ODB dependencies as much as possible

PR #1016, opened 2020-02-21
Fix bug in microschema migration

philippguertler:
## Abstract
Fixed a bug that caused errors during microschema migrations in certain situations.
### General

### On API Changes


cschockaert:
love you, we ran into this bug this morning, a binary field node caused a NPE during migration !

Jotschi:
@philippguertler Please rebase this on hotfix-1.4.x since we want this also in 1.4.x

PR #1018, opened 2020-02-24
Remove more ferma code

Jotschi:
## Abstract

Reduces ferma code and adds test for `madl-orientdb` module.

### General

### On API Changes


madl/orientdb/src/main/java/com/gentics/madl/ElementTypeClassCache.java
```diff
@@ -0,0 +1,37 @@
...
+import org.reflections.Reflections;
+
+import com.gentics.madl.annotations.GraphElement;
+
+/**
+ * Type cache which also provides resolving methods which cache the result.
+ */
+public class ElementTypeClassCache {
+
+	private final Map<String, Class> classStringCache = new HashMap<>();
```
philippguertler: ConcurrentHashmap

madl/orientdb/src/main/java/com/gentics/madl/ext/orientdb/OrientDBTypeResolver.java
```diff
@@ -0,0 +1,95 @@
...
+import com.syncleus.ferma.traversals.VertexTraversal;
+import com.syncleus.ferma.typeresolvers.TypeResolver;
+import com.tinkerpop.blueprints.Element;
+import com.tinkerpop.blueprints.impls.orient.OrientEdge;
+import com.tinkerpop.blueprints.impls.orient.OrientVertex;
+import com.tinkerpop.blueprints.util.wrappers.wrapped.WrappedElement;
+
+public class OrientDBTypeResolver implements TypeResolver {
+
+	private final ElementTypeClassCache elementTypCache;
```
philippguertler: Rename to elementTypeCache

Jotschi:
@philippguertler I checked the performance.

I compared loading via loading via `ov.getType().getName()` (suggested new resolver) versus 
`ov.getProperty("fermaType")` (current way).

I ran the resolver test 50 times each on 1.5 Million vertices to get an average value.

The new implementation is 30% to 23% faster.

Used test: https://gist.github.com/Jotschi/e1fd6ae5ae9740f2fff7b663fb1ae08f

PR #1021, opened 2020-02-27
Add branch deletion

philippguertler:
## Abstract
Allows the deletion of branches which are not latest.

### General

### On API Changes


rest-model/src/main/java/com/gentics/mesh/core/rest/admin/consistency/InconsistencyInfo.java
```diff
@@ -26,7 +26,7 @@
 	private boolean repaired = false;
 
 	@JsonProperty(required = true)
-	@JsonPropertyDescription("Repair action which will attept to fix the inconsistency. The action will only be invoked when using invoking the rapair endpoint.")
+	@JsonPropertyDescription("Repair action which will attempt to fix the inconsistency. The action will only be invoked when using invoking the repair endpoint.")
```
Jotschi: :rofl:

PR #1036, opened 2020-03-16
Fix "thread blocked" during startup

mephinet:
## Abstract

Moving slow operations during startup into worker threads. Tested using `cpulimit -l 30`.

### General


philippguertler:
buildnow

PR #1039, opened 2020-03-19
Ferma Code Reduction

Jotschi:
## Abstract

Reduce more ferma code

### General

### On API Changes


madl/madl-ferma/src/main/java/com/gentics/madl/frame/AbstractVertexFrame.java
```diff
@@ -146,12 +154,30 @@ public void setSingleLinkOutTo(VertexFrame vertex, String... labels) {
...
+
 	@Override
 	public <T extends ElementFrame> TraversalResult<? extends T> in(String label, Class<T> clazz) {
 		TraversalResult<? extends T> result = new TraversalResult<>(in(label).frameExplicit(clazz));
 		return result;
 	}
 
+	@Override
+	public Stream<Edge> streamInE(String label) {
+		Iterator<Edge> it = getElement().getEdges(Direction.IN, label).iterator();
```
philippguertler: Create a private function which takes the direction as an argument to avoid code duplication. Maybe even consider making it public.

Jotschi: Yes, I was wondering myself. I try to add those methods to the interface via default methods. This way all implementations can use those generic methods which make use of `getElement`.
WDYT

Jotschi: @philippguertler  done

philippguertler: Use the new function for streamOutE as well

Jotschi: Done

Jotschi:
No longer needed

PR #1055, opened 2020-04-17
Auto feature image manipulation hotfix 1.4.x

dajenacuko:
## Abstract
Mesh supports now the auto value in height or width. The changes are done in the Image Manipulation and in BinaryFieldResponse. The height and width are now strings.
### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryFieldResponseHandler.java
+++ b/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryFieldResponseHandler.java
@@ -1,5 +1,12 @@
 package com.gentics.mesh.core.endpoint.node;
 
+import static com.gentics.mesh.http.HttpConstants.ETAG;
+import static com.gentics.mesh.util.MimeTypeUtils.DEFAULT_BINARY_MIME_TYPE;
+import static io.netty.handler.codec.http.HttpResponseStatus.NOT_MODIFIED;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
 import com.gentics.mesh.context.InternalActionContext;
 import com.gentics.mesh.context.impl.InternalRoutingActionContextImpl;
 import com.gentics.mesh.core.data.binary.Binary;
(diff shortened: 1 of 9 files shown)


services/image-imgscalr/src/main/java/com/gentics/mesh/image/ImgscalrImageManipulator.java
```diff
@@ -206,13 +215,11 @@ private ImageReader getImageReader(ImageInputStream input) {
 
 	/**
 	 * Create an image writer from the same image format as the specified image reader writing to the given output stream.
-	 *
+	 * <p>
```
Jotschi, line 218:
```suggestion
```

core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryFieldResponseHandler.java
```diff
@@ -130,22 +129,32 @@ private void resizeAndRespond(RoutingContext rc, BinaryGraphField binaryField, I
...
+					response.putHeader(MeshHeaders.WEBROOT_RESPONSE_TYPE, "binary");
+					// Set to IDENTITY to avoid gzip compression
+					response.putHeader(HttpHeaders.CONTENT_ENCODING, HttpHeaders.IDENTITY);
+
+					addContentDispositionHeader(response, fileName, "inline");
+
+					response.sendFile(cachedFilePath);
+				}))
+			.subscribe(ignore -> {
+			}, rc::fail);
```
Jotschi, line 157:
```suggestion
			.subscribe(ignore -> {}, rc::fail);
```

rest-model/src/main/java/com/gentics/mesh/parameter/ImageManipulationParameters.java
```diff
@@ -26,67 +26,96 @@
...
+
+	/**
+	 * Set the image width.
+	 *
+	 * @param width
+	 * @return Fluent API
+	 * @deprecated Use {@link #setWidth(String)} instead.
+	 */
+	@Deprecated
 	default ImageManipulationParameters setWidth(Integer width) {
```
Jotschi, line 64: I think both methods are valid. It is bad to suggest a dev to transform his height values to string. Let us do this for him. Thus keep the default method and don't deprecate it.

PR #1056, opened 2020-04-17
Include all previous schema versions when migrating remaining nodes 

philippguertler:
## Abstract
When migrating, currently only schemas that are active to the branch are considered for migration. Under certain circumstances, other older nodes can exist. This fix ensures that these older Nodes are migrated as well.

### General

### On API Changes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -32,6 +32,8 @@ icon:plus[] Server: The `serverTokens` flag and `MESH_HTTP_SERVER_TOKENS` env ha
 
 icon:check[] Changelog: The automated recovery from v1.4.4 now renames duplicate versions instead of deleting them.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.adoc
```diff
@@ -32,6 +32,8 @@ icon:plus[] Server: The `serverTokens` flag and `MESH_HTTP_SERVER_TOKENS` env ha
 
 icon:check[] Changelog: The automated recovery from v1.4.4 now renames duplicate versions instead of deleting them.
 
+icon:check[] Core: Migrating remaining nodes via the `/{project}/branches/{branchUuid}/migrateSchemas` and `/{project}/branches/{branchUuid}/migrateMicroschemas` now includes all previous schema versions instead of only active schema versions.
```
Jotschi, line 35: Would it be much effort to also add this to migrateSchemas?

PR #1059, opened 2020-04-20
Add perm checks for server tokens

Jotschi:
## Abstract

Add perm checks for server tokens to always allow admin users to see the tokens.

### General

### On API Changes

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -28,6 +28,8 @@ include::content/docs/variables.adoc-include[]
 [[v1.4.7]]
 == 1.4.7 (TBD)
 
+icon:plus[] Server: The `/api/v1/RAML` endpoint can now only be accessed by admin users.
+
 icon:plus[] Server: Admin users can now always see the `serverTokens` regardless of the `serverTokens` setting.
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.adoc
```diff
@@ -28,6 +28,10 @@ include::content/docs/variables.adoc-include[]
 [[v1.4.7]]
 == 1.4.7 (TBD)
 
+icon:plus[] Server: The `/api/v1/RAML` endpoint can now only be accessed by admin users.
```
philippguertler, line 31:
```suggestion
icon:plus[] Server: The `/api/v2/raml` endpoint can now only be accessed by admin users.
```

core/src/main/java/com/gentics/mesh/core/endpoint/handler/AbstractHandler.java
```diff
@@ -22,4 +24,5 @@ protected void validateParameter(String value, String name) {
 			throw error(BAD_REQUEST, "error_request_parameter_missing", name);
 		}
 	}
+
```
philippguertler, line 27: Nothing has changed here

PR #1065, opened 2020-04-23
Refactor plugin init handling

Jotschi:
## Abstract

Changes plugin deployment process. Plugins will now pre-registered and registered once mesh is ready.

### General

### On API Changes


rest-model/src/main/java/com/gentics/mesh/core/rest/plugin/PluginResponse.java
```diff
@@ -22,6 +22,10 @@
 	@JsonPropertyDescription("Manifest of the plugin")
 	private PluginManifest manifest;
 
+	@JsonProperty(required = true)
+	@JsonPropertyDescription("Status of the plugin")
+	private PluginStatus status;
+
 	public PluginResponse() {
```
Jotschi, line 29: Maybe we should include the potential error message on failures (maybe a security concern although admin perm is needed)
Otherwise we could include the last successful or the error phase in which the plugin failed.

@mephinet

distributed/src/test/java/com/gentics/mesh/distributed/PluginClusterTest.java
```diff
@@ -0,0 +1,131 @@
...
+	 * Check whether the plugin can be found and is registered. Otherwise fail after the given timeout is exceeded.
+	 * 
+	 * @param container
+	 * @param timeoutInMilliseconds
+	 * @throws InterruptedException
+	 */
+	private void waitForPluginRegistration(MeshDockerServer container, int timeoutInMilliseconds) throws InterruptedException {
+		long start = System.currentTimeMillis();
+		PluginResponse plugin = null;
+		while (true) {
```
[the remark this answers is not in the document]
Jotschi: Not possible since this is not embedded. Using websocket would be error prone since the docker container is started async this way we can't 100% ensure that we catch the event of the started container.

core/src/main/java/com/gentics/mesh/plugin/registry/DelegatingPluginRegistryImpl.java
```diff
@@ -0,0 +1,188 @@
...
+		this.vertx = vertx;
+		this.db = db;
+		this.pluginLock = pluginLock;
+	}
+
+	@Override
+	public void start() {
+		// Check to avoid bogus restarts
+		if (timerId == -1) {
+			timerId = vertx.get().setPeriodic(1000, id -> register());
```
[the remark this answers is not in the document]
Jotschi: Lets check this again with the next review.

core/src/test/resources/logback-test.xml
```diff
@@ -44,7 +44,7 @@
 	<logger name="com.gentics.mesh.search.verticle" level="TRACE" />
 	<logger name="com.gentics.mesh.plugin" level="TRACE" />
  -->
-	<root level="ERROR">
+	<root level="DEBUG">
```
Jotschi, line 47: Revert to error

core/src/main/java/com/gentics/mesh/plugin/registry/DelegatingPluginRegistryImpl.java
```diff
@@ -0,0 +1,204 @@
...
+		this.authServiceRegistry = authServiceRegistry;
+		this.manager = manager;
+		this.rxVertx = rxVertx;
+		this.db = db;
+	}
+
+	@Override
+	public void start() {
+		// EventBus eb = rxVertx.get().eventBus();
+		log.debug("Starting to listen to plugin pre-registered events");
```
philippguertler, line 76: There is something missing or the log message is wrong

core/src/main/java/com/gentics/mesh/plugin/registry/DelegatingPluginRegistryImpl.java
```diff
@@ -0,0 +1,204 @@
...
+		}
+	}
+
+	@Override
+	public void preRegister(MeshPlugin plugin) {
+		EventBus eb = rxVertx.get().eventBus();
+		Objects.requireNonNull(plugin, "The plugin must not be null");
+		manager.get().setStatus(plugin.id(), PluginStatus.PRE_REGISTERED);
+		// TODO add payload
+		eb.publish(MeshEvent.PLUGIN_PRE_REGISTERED.getAddress(), null);
```
philippguertler, line 113: Payload is still missing, if it's not difficult, we should add it now

core/src/main/java/com/gentics/mesh/plugin/registry/DelegatingPluginRegistryImpl.java
```diff
@@ -0,0 +1,204 @@
...
+
+	/**
+	 * Use the lock in clustered mode to prevent concurrent inits
+	 * 
+	 * @param lockedAction
+	 * @return
+	 */
+	private Completable optionalLock(Completable lockedAction) {
+		if (options.getClusterOptions().isEnabled()) {
+			return rxVertx.get().sharedData().rxGetLockWithTimeout(GLOBAL_PLUGIN_LOCK_KEY, 10_000).toMaybe()
```
philippguertler, line 174: This should ideally be configurable or at least be derived from the plugin initialize timeout

PR #1066, opened 2020-04-28
Search analyzer per language

philippguertler:
## Abstract
Allows specifying Elasticsearch settings / mappings per language

Resolves https://github.com/gentics/mesh-incubator/issues/218
Specification: https://github.com/gentics/mesh-incubator/issues/218#issuecomment-617762317
### General

### On API Changes


doc/src/main/docs/elasticsearch.asciidoc
```diff
@@ -684,6 +684,87 @@ TIP: You can use the `POST {apiLatest}/utilities/validateSchema` endpoint to val
...
 
+[[language-override]]
+=== Mappings / Settings per Language
+
+It is also possible to define different settings / mappings for different languages. To do so, add the `_meshLanguageOverride` property to the `elasticsearch` object.
+In this object you can override settings / mappings for each language. You can also provide the same setting / mapping for multiple language by using a
+comma separated list of languages as the key.
+
+For each overridden language, Gentics Mesh will create additional indices in Elasticsearch. Node contents of that language will be saved in that index.
+When a node content is created with a language that is not overridden, the node content will be saved in the default index. The settings / mappings of the
```
philippguertler, line 695:
```suggestion
comma separated list of languages as the key.

NOTE: For each overridden language, Gentics Mesh will create additional indices in Elasticsearch. This might affect the resources used by Elasticsearch.

 Node contents of that language will be saved in that index.
When a node content is created with a language that is not overridden, the node content will be saved in the default index. The settings / mappings of the
```

PR #1069, opened 2020-05-04
Drop all node indices when project is deleted

philippguertler:
## Abstract
Language specific indices were not deleted

### General

### On API Changes


philippguertler:
Changelog is not necessary because this hasn't been released yet

PR #1078, opened 2020-05-12
Allow empty binaries without inconsistency

philippguertler:
## Abstract
Uploading a binary with size 0 should not cause an inconsistency in the `/admin/consistency/check` endpoint.

### General

### On API Changes


cschockaert:
❤️

PR #1079, opened 2020-05-12
Add fix for plugin undeploy NPE

Jotschi:
## Abstract

* Fixes NPE in undeployment code
* Adds more logging on cluster lock handling
* Re-order lock and quorum check during plugin deployments

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
@@ -278,6 +278,7 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 			resolvePlugins();
 			PluginWrapper wrapper = getPlugin(id);
 			if (wrapper == null) {
+				log.debug("Plugin with id {} could not be found", id);
 				return;
 			}
 			Plugin plugin = wrapper.getPlugin();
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -277,6 +277,9 @@ public Completable undeploy(String id) {
 		return Completable.fromRunnable(() -> {
 			resolvePlugins();
 			PluginWrapper wrapper = getPlugin(id);
+			if (wrapper == null) {
+				return;
```
philippguertler, line 281: Maybe output a log message in this case

core/src/main/java/com/gentics/mesh/plugin/registry/DelegatingPluginRegistryImpl.java
```diff
@@ -161,16 +161,20 @@ private Completable registerAndInitalizePlugin(MeshPlugin plugin) {
...
 	 * @return
 	 */
-	private Completable optionalLock(Completable lockedAction) {
+	private Completable optionalLock(Completable lockedAction, String id) {
 		int timeout = options.getPluginTimeout();
 		if (options.getClusterOptions().isEnabled()) {
 			return rxVertx.get().sharedData().rxGetLockWithTimeout(GLOBAL_PLUGIN_LOCK_KEY, timeout * 2 * 1000).toMaybe()
 				.flatMapCompletable(lock -> {
-					log.debug("Acquired lock for plugin registration.");
+					log.info("Acquired lock for registration of plugin {}", id);
```
philippguertler, line 172: Why info? I would leave it at debug

Issue #975, opened 2019-12-18
Assert allow field during node field updates


Jotschi:
I noticed that the allow field will currently not be validated strictly in the API.

We should add the same checks that have also been added via #431

karowin:
HI, @Jotschi can you please add some more details to this, like steps to reproduce and test this.

karowin:
HI, @Jotschi could you please respond to my query above.

Jotschi:
@karowin 

I quickly sketched a test that should help you. The test should assert for the `…NotAllowed` cases that an error is thrown when a node field gets updated with a node that uses a different schema compared to the allowed value in the schema of the node that gets updated.
The test currently passes and expects a pass - you would need to change the assertion for this case.

You can use the call function to expect a specific error. Example:
```java
call(() -> client().createNode(PROJECT_NAME, request), BAD_REQUEST, "node_no_languagecode_specified");
```
I assume a new I18n entry also needs to be added with a meaningful message in case of failing allow checks.


Besides the test there are two places which need to be updated:

1. Similar to this check a check on the node's schema name needs to be performed against the provided `schema`
* https://github.com/gentics/mesh/blob/dev/core/src/main/java/com/gentics/mesh/core/data/node/field/impl/NodeGraphFieldImpl.java#L81

2. For lists this has to be checked for each individual entry:
https://github.com/gentics/mesh/blob/dev/core/src/main/java/com/gentics/mesh/core/data/node/field/list/impl/NodeGraphFieldListImpl.java#L88


```java
package com.gentics.mesh.core.schema;

import static com.gentics.mesh.test.TestDataProvider.PROJECT_NAME;
import static com.gentics.mesh.test.TestSize.FULL;
import static com.gentics.mesh.test.context.ElasticsearchTestMode.TRACKING;

import java.util.Collections;

import org.junit.Before;
import org.junit.Test;

import com.gentics.mesh.core.rest.node.FieldMapImpl;
import com.gentics.mesh.core.rest.node.NodeCreateRequest;
import com.gentics.mesh.core.rest.node.field.Field;
import com.gentics.mesh.core.rest.node.field.impl.NodeFieldImpl;
import com.gentics.mesh.core.rest.node.field.list.impl.NodeFieldListImpl;
import com.gentics.mesh.core.rest.node.field.list.impl.NodeFieldListItemImpl;
import com.gentics.mesh.core.rest.schema.FieldSchema;
[...]

PR #1082, opened 2020-05-25
Fix allow field node update

karowin:
### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/node/field/impl/NodeGraphFieldImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/node/field/impl/NodeGraphFieldImpl.java
@@ -100,7 +100,7 @@ public class NodeGraphFieldImpl extends MeshEdgeImpl implements NodeGraphField {
 			// TODO check whether we want to allow cross project node references
 
 			NodeFieldSchema nodeFieldSchema = (NodeFieldSchema) fieldSchema;
-			String schemaName = schema.getName();
+			String schemaName = node.getSchemaContainer().getName();
 
 			if (!ArrayUtils.isEmpty(nodeFieldSchema.getAllowedSchemas())
 					&& !Arrays.asList(nodeFieldSchema.getAllowedSchemas()).contains(schemaName)) {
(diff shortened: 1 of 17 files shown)


core/src/main/java/com/gentics/mesh/core/data/node/field/impl/NodeGraphFieldImpl.java
```diff
@@ -95,6 +98,17 @@
 			// Check whether the container already contains a node field
 			// TODO check node permissions
 			// TODO check whether we want to allow cross project node references
+
+			NodeFieldSchema nodeFieldSchema = (NodeFieldSchema) fieldSchema;
+			String schemaName = schema.getName();
```
Jotschi, line 103: I think this is not correct. The name of the schema should be taken from the node.
E.g. node.getSchema().getName()

karowin: HI @Jotschi , thanks for the comment. I did use String schemaName = node.getSchemaContainer().getName(); but with this, I always get the schema name as "folder" and my positive test cases fails because of this.
_org.junit.ComparisonFailure: 
Expected :The value for field "testField" must not use schema "test".
Actual   :The value for field "testField" must not use schema "folder"._

Jotschi: Ok. Strange.I need to take a closer look and debug it inline. I'm however currently unable to do so. Maybe @philippguertler can give you feedback on this.

philippguertler: @Jotschi is correct. You must use the schema of the referenced node to check here. It seems that the tests are wrong if this results in an error.

core/src/test/java/com/gentics/mesh/core/schema/SchemaAllowNodeFieldTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private void runTestAndExpectFailure(FieldSchema schemaField, Field nodeField) {
+        createSchema(schemaField);
+        createNodeAndExpectFailure(nodeField);
+    }
+
+    @Test
+    public void node() {
+        runTest(
+                new NodeFieldSchemaImpl().setAllowedSchemas("test"),
+                new NodeFieldImpl().setUuid(nodeUuid));
```
philippguertler, line 85: The referenced node is a folder. `nodeUuid` is set in the `setup()` method, where you can see that this is a folder. You need to create a node of the `test` schema and use that UUID here. Same for the other tests in this class.

karowin: Thanks @philippguertler , I have updated it.

philippguertler:
buildnow

philippguertler:
Fixes #975

PR #1103, opened 2020-07-30
MDM Basics

philippguertler:
Basic interfaces and wrappers for `UserRoot` and `User`


philippguertler:
Will be done in a different method, with less overall changes at once.

PR #1128, opened 2020-09-21
Add liveness probe check for failed plugins

Jotschi:
## Abstract

Adds liveness probe checks for failed plugins. This way K8S will restart the pod whenever a plugin deployment fails.

### General

### On API Changes


philippguertler (changes requested):
Use warn for logging

philippguertler (approved):

PR #1145, opened 2020-10-08
Add cluster changelog check

Jotschi:
## Abstract

This change prevents nodes from joining a cluster which has an outdated database. The changelog should not be executed by joining members to prevent concurrency issues during the startup process.

This fix addresses the problem of upgrading cluster members without applying the changelog.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/cli/BootstrapInitializerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/cli/BootstrapInitializerImpl.java
@@ -250,6 +250,14 @@ public class BootstrapInitializerImpl implements BootstrapInitializer {
 				// Only execute the changelog if there are any elements in the graph
 				invokeChangelog();
 			}
+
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/gentics/mesh/cli/BootstrapInitializerImpl.java
```diff
@@ -250,6 +250,14 @@ private boolean initLocalData(MeshOptions configuration, boolean isJoiningCluste
 				// Only execute the changelog if there are any elements in the graph
 				invokeChangelog();
 			}
+
+			if (isJoiningCluster && requiresChangelog()) {
```
philippguertler, line 254: How can you guarantee that the changes have been replicated from another instance?

Jotschi: `db.clusterManager().start();` will block during activation of the server until the database has been synced.

PR #1153, opened 2020-10-30
UVEK-455 Add versioning flag to schemas

yrucrem:
## Abstract

Versioning can be disabled per Schema by setting the new versioning flag to false.

### General

### On API Changes


common/src/main/java/com/gentics/mesh/core/data/schema/SchemaContainerVersion.java
```diff
@@ -60,9 +60,14 @@ default TypeInfo getTypeInfo() {
 	TraversalResult<? extends Node> getNodes(String branchUuid, User user, ContainerType type);
 
 	/**
-	 * Check whether versioning is disabled by default or via the schema setting.
+	 * Check whether auto purge is enabled by default or via the schema setting.
```
Jotschi, line 63: Revert this

PR #1157, opened 2020-12-02
Hotfix 1.4.x pdf metadata

plyhun:
## Abstract

The binary file with XML-based metadata (PDF) may be refused to upload into Mesh because of an array metadata field key. The provided fix allows such files for an upload. 

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/node/field/impl/BinaryGraphFieldImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/node/field/impl/BinaryGraphFieldImpl.java
@@ -14,8 +14,6 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import org.apache.commons.lang.StringUtils;
-
 import com.gentics.madl.index.IndexHandler;
 import com.gentics.madl.tx.Tx;
 import com.gentics.madl.type.TypeHandler;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/gentics/mesh/core/data/node/field/impl/BinaryGraphFieldImpl.java
```diff
@@ -331,17 +331,18 @@ public Binary getBinary() {
...
-		return metadata;
-	}
+    public Map<String, String> getMetadataProperties() {
+        List<String> keys = getPropertyKeys().stream().filter(k -> k.startsWith(META_DATA_PROPERTY_PREFIX)).collect(Collectors.toList());
+        Map<String, String> metadata = new HashMap<>();
+        for (String key : keys) {
+            String name = key.substring(META_DATA_PROPERTY_PREFIX.length());
+            name = key.replaceAll("%5B", "[");
+            name = key.replaceAll("%5D", "]");
+            String value = property(key);
```
Jotschi, line 341:
```suggestion
            String value = property(name);
```

core/src/main/java/com/gentics/mesh/core/data/node/field/impl/BinaryGraphFieldImpl.java
```diff
@@ -331,17 +331,18 @@ public Binary getBinary() {
...
-		}
-		return metadata;
-	}
+    public Map<String, String> getMetadataProperties() {
+        List<String> keys = getPropertyKeys().stream().filter(k -> k.startsWith(META_DATA_PROPERTY_PREFIX)).collect(Collectors.toList());
+        Map<String, String> metadata = new HashMap<>();
+        for (String key : keys) {
+            String name = key.substring(META_DATA_PROPERTY_PREFIX.length());
+            name = key.replaceAll("%5B", "[");
+            name = key.replaceAll("%5D", "]");
```
Jotschi, line 340:
```suggestion
            name = name.replaceAll("%5D", "]");
```

Jotschi (changes requested):

Jotschi:
Please also use the correct github email address in your commits. Otherwise the commits can't be tracked to your user. Use git ammend / rebase when you fixed the setting.

plyhun:
Done.

PR #1159, opened 2020-12-15
Properly fetch bucket-id

micstra:
## Abstract
The bucket-id-property is stored as a long wrapper-type.
This leads to a ClassCastException if just the plain getProperty is returned.

Fetch the property as long and then manually convert to an Integer to avoid ClassCastExceptions.
### General


Jotschi:
Thanks for catching this. I have modified the change and pushed it into dev. It will be included in the next release.

PR #1162, opened 2020-12-22
Split GraphDB-related options

plyhun:
## Abstract
Extract the DB-independent options to an abstract class.

**BREAKING:** MeshOptions is now abstract and cannot be instantiated!

### General

### On API Changes


api/src/main/java/com/gentics/mesh/etc/config/OrientDBMeshOptions.java
```diff
@@ -0,0 +1,53 @@
...
+			getStorageOptions().validate(this);
+		}
+		if (getGraphQLOptions() != null) {
+			getGraphQLOptions().validate(this);
+		}
+		Objects.requireNonNull(getNodeName(), "The node name must be specified.");
+		if (getVersionPurgeMaxBatchSize() <= 0) {
+			throw new IllegalArgumentException("versionPurgeMaxBatchSize must be positive.");
+		}
+		// TODO check for other invalid characters in node name
```
Jotschi, line 45: super.validate()

plyhun: Done.

api/src/main/java/com/gentics/mesh/etc/config/MeshOptions.java
```diff
@@ -513,9 +496,6 @@ public void validate() {
 		if (getContentOptions() != null) {
 			getContentOptions().validate(this);
 		}
-		if (getGraphQLOptions() != null) {
```
plyhun, line 516: Get this back!

PR #1164, opened 2021-01-27
CL-696 bootstrapping

deckdom:
Abstraction of the BoostrapInitializer and other Components


common/src/main/java/com/gentics/mesh/core/link/WebRootLinkReplacerImpl.java
```diff
@@ -11,6 +11,7 @@
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.gentics.mesh.cli.ODBBootstrapInitializer;
```
npomaroli, line 14: Check imports:
* Maintain order of imports
* Do not use wildcard imports
* Static imports come first

core/src/main/java/com/gentics/mesh/dagger/module/CommonBindModule.java
```diff
@@ -135,7 +136,10 @@
 	abstract DropIndexHandler bindCommonHandler(DropIndexHandlerImpl e);
 
 	@Binds
-	abstract BootstrapInitializer bindBoot(BootstrapInitializerImpl e);
+	abstract BootstrapInitializer bindBoot(ODBBootstrapInitializerImpl e);
```
npomaroli, line 139: Is it possible to move all references to ODB... classes out of the CommonBindModule?

plyhun: Technically yes, with providing a level of abstraction to the common binder, with a concrete impl in the context-specific binder. See https://github.com/gentics/mesh/pull/1162/files#diff-c25726c66cc9f59426bc6b4f8237845e8b51685bb3efb86ebd3fab0e246ccba9L194 and https://github.com/gentics/mesh/pull/1162/files#diff-3728767650796dea65e91cd45cde6cdb0d6bc46aad5b0c107dd0aa29dbcd4beaR92 .
Since @deckdom is busy, I will make this change.

PR #1165, opened 2021-02-05
Generalize DAOs into MDM API

plyhun:
Extract DB-agnostic parts into MDM API.

Actions performed:
1. Create missing MDM API entities:
* DAOs (with Hib* prefix, matching the common DAO naming)
* DB-agnostic Tx
* Fields
2. Replace *DaoWrapper with *Dao where possible
3. Delete *DaoWrappers
* if the wrapper is empty (fully transferred into MDM API), just delete the class and its references
* if the wrapper contains non-transferrable elements, rename it into OrientDB*Dao
4. Fix the references in the main code
5. Refactor the code, which relies on GraphDB / OrientDB API
6. Fix the references in the unit test code

TODO
1.Check failing/unstable tests
2. Document the classes, where required


plyhun:
Abandoned in favor of Unittest-aware https://github.com/gentics/mesh/pull/1224

PR #1174, opened 2021-03-22
Improve plugin client endpoint creation for IP = 0.0.0.0

plyhun:
## Abstract

The plugin environment crashes on deployment the plugin, that utilizes non-admin REST client, if the host OS does not support serving over address 0.0.0.0.

## Checklist

* [x] Fall back the 0.0.0.0 into 127.0.0.1

### General

### On API Changes


bernhardriegler (approved):

yrucrem (commented):
The fix should go to 1.4.x, otherwise it looks good to me.

PR #1182, opened 2021-03-29
feature/MS3-5 Add S3 binary field

rhoxhaj:
## Abstract
Code for the creation of the S3 binary field. This branch covers only the S3 binary field, without tests and without extended functionality that will be covered from future tickets (This is the ticket board: https://jira.gentics.com/secure/RapidBoard.jspa?rapidView=103&projectKey=MS3&view=detail)


plyhun:
So far so good. Need the actual implementation and the corresponding tests though.

PR #1183, opened 2021-03-30
Extend configuration for AWS S3 usage

rhoxhaj:
## Abstract
Extend configuration for AWS in yml and configuration files. There are 2 main configurations, one for S3 and another one for S3 Cache. The nested yml structure is reflected into the implemented classes.


api/src/main/java/com/gentics/mesh/etc/config/MeshOptions.java
```diff
@@ -211,6 +215,17 @@ public MeshOptions setStorageOptions(GraphStorageOptions storageOptions) {
 		return this;
 	}
 
+	@JsonProperty("s3binary")
+	public S3Options getS3Options() {
```
yrucrem, line 219: Even though the `S3Options` do not have a validation method right now, I would still add them in the `MeshOptions` validate() method.

api/src/main/java/com/gentics/mesh/etc/config/S3Options.java
```diff
@@ -0,0 +1,91 @@
...
+    @EnvironmentVariable(name = MESH_S3_BINARY_SECRET_ENV, description = "Override the configured AWS S3 secret.")
+    private String secret;
+
+    @JsonProperty(required = true)
+    @JsonPropertyDescription("S3 Cache Bucket Options.")
+    private S3CacheOptions s3cacheOptions = new S3CacheOptions();
+
+    @JsonProperty("cache")
+    public S3CacheOptions getS3CacheOptions() {
+        return this.s3cacheOptions;
```
yrucrem, line 83: Make sure this does not return `null`.

yrucrem (changes requested):

yrucrem (approved):

plyhun (approved):

PR #1185, opened 2021-04-06
Feature/ms3 7 upload s3

rhoxhaj:
## Abstract
Add Upload functionality by creating a new REST endpoint. This endpoint should:

1. create a presigned url for the requested node + fieldName.
2. save to Mesh the reference (AWS object key).
3. return the URL with the necessary parameters and method.

Additional Info:
- The AWS SDK had to be updated in order to include the latest AWS functionalities, like presigned URLs.
- The tests are to be covered in a separate ticket.
- There are some class variables (mostly metadata ones) that are not removed, since in the future those will be needed for metadata extraction.


common/src/main/resources/i18n/translations_zh.properties
```diff
@@ -21,6 +21,7 @@ error_element_for_document_type_not_found=找不到文档类型为“{1}”且uu
 error_binaryfield_not_found_with_name=找不到二进制字段“{0}”。
 error_field_not_found_with_name=找不到字段“{0}”。
 error_found_field_is_not_binary=字段“{0}”的数据模型不是二进制字段模型。
+error_found_field_is_not_s3_binary=字段“{0}”的数据模型不是二进制字段模型。
```
plyhun, line 24: I see the value being copied from plain binary message. Since no one of us speaks Chinese, I would suggest getting the English version through translator service, and paste the results here (until a native Chinese speaker proposes a better translation).

rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/S3BinaryField.java
```diff
@@ -11,22 +11,22 @@
...
-    S3BinaryField setS3BinaryUuid(String uuid);
+    S3BinaryField setS3binaryUuid(String uuid);
 
     /**
      * Return the s3binary filesize.
      *
      * @return Filesize in bytes
      */
-    long getFileSize();
+    Long getFileSize();
```
plyhun, line 29: Why use boxing here?

common/src/main/resources/i18n/translations_zh.properties
rhoxhaj: You can check this again.

rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/S3BinaryField.java
rhoxhaj: I am using it here since filesize is not required to be inserted in the Mesh immediately and by using a long primitive, i was actually inserting a field filesize=0. I saw that in many fieldtypes the Number Classes are preferred. Let me know if I should change it.

services/aws-s3-storage/src/test/java/com/gentics/mesh/storage/s3/S3BinaryStorageTest.java
```diff
@@ -56,8 +58,8 @@ public void testStore() {
 		Mockito.when(mockField.getBinary()).thenReturn(binary);
 		Mockito.when(binary.getSHA512Sum()).thenReturn("test");
 //		assertFalse(storage.exists(mockField));
-		storage.store(Flowable.just(Buffer.buffer("test")), "test").blockingAwait();
-		assertTrue(storage.exists(mockField));
+//		storage.store(Flowable.just(Buffer.buffer("test")), "test").blockingAwait();
+//		assertTrue(storage.exists(mockField));
```
plyhun, line 62: I assume this test to be extended, right?

rhoxhaj: Yes, tests are covered in a separate ticket.

PR #1188, opened 2021-04-15
Feature/ms 10 download s3

rhoxhaj:
## Abstract
Adapt endpoints used to get binary data:

    https://getmesh.io/docs/api/#project__webrootfield__fieldName___path__get
    https://getmesh.io/docs/api/#project__webroot__path__get
    https://getmesh.io/docs/api/#project__nodes__nodeUuid__binary__fieldName__get

In each case, the response should be a redirect 302 with the presigned URL. There should be possible also to manipulate images when requested via the parameters.


core/src/main/java/com/gentics/mesh/core/data/container/impl/NodeGraphFieldContainerImpl.java
```diff
@@ -724,10 +722,15 @@ public String getSegmentFieldValue() {
...
 		if (stringField == null) {
-			BinaryGraphField binaryField = getBinary(segmentFieldKey);
-			if (binaryField != null) {
-				return binaryField.getFileName();
+			S3BinaryGraphField s3binaryField = getS3Binary(segmentFieldKey);
+			if (nonNull(s3binaryField) && nonNull(s3binaryField.getS3Binary())) {
+				return s3binaryField.getS3Binary().getFileName();
 			}
+			BinaryGraphField binary = getBinary(segmentFieldKey);
+			if (nonNull(binary) && nonNull(binary.getBinary())) {
```
plyhun, line 730: Are you sure regarding the extra check for `binary.getBinary()`?

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -1985,6 +1986,22 @@ private PathSegment getSegment(String branchUuid, ContainerType type, String seg
 					return new PathSegmentImpl(container, binaryField, container.getLanguageTag(), segment);
 				}
 			}
+			//TODO S3
```
plyhun, line 1989: Is there something left here to do?

core/src/main/java/com/gentics/mesh/core/endpoint/node/S3UploadContext.java
```diff
@@ -30,6 +30,7 @@ public void setS3ObjectKey(String s3ObjectKey) {
 		this.s3ObjectKey = s3ObjectKey;
 	}
```
plyhun, line 32: This line change may be reverted.

core/src/main/java/com/gentics/mesh/core/endpoint/webroot/WebRootHandler.java
```diff
@@ -98,7 +103,13 @@ public void handleGetPath(RoutingContext rc) {
 				}
 				binaryFieldResponseHandler.handle(rc, binaryField);
 				return null;
```
plyhun, line 105: Here may be `else if`.

mdm/common/src/main/java/com/gentics/mesh/core/image/ImageManipulator.java
```diff
@@ -1,60 +1,70 @@
...
+import java.awt.image.BufferedImage;
+import java.io.InputStream;
+import java.util.Map;
+
 /**
  * SPI provider interface for image manipulators.
  */
 public interface ImageManipulator {
 
-	/**
```
plyhun, line 17: Something happened to this file's indentation - it looks like unchanged code-wise.

plyhun:
Except abovementioned, all looks ok. Before merging the whole S3 into the target release branch I will anyway be checking everything for working and passing the unit tests.

core/src/main/java/com/gentics/mesh/core/data/container/impl/NodeGraphFieldContainerImpl.java
rhoxhaj: Actually I found this relevant for S3BinaryField but if binary.getBinary() has worked for binary fields, it is better to revert this change. I will revert it and check only if this is not null.

core/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
rhoxhaj: No, I had forgot to remove this. Done now :)

PR #1189, opened 2021-04-20
Feature/ms3 11 adapt binary transformation

rhoxhaj:
## Abstract
when this endpoint is called with a s3binary field:

- Mesh will download the binary (in the temp folder)
- apply transformations (same as existing binaryTransformation) and return json with updated information.


rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/S3BinaryFieldTransformRequest.java
```diff
@@ -0,0 +1,234 @@
...
+import com.gentics.mesh.core.rest.common.RestModel;
+import com.gentics.mesh.core.rest.node.field.image.FocalPoint;
+import com.gentics.mesh.parameter.image.CropMode;
+import com.gentics.mesh.parameter.image.ImageRect;
+import com.gentics.mesh.parameter.image.ResizeMode;
+
+/**
+ * POJO for a binary field transform request
+ */
+public class S3BinaryFieldTransformRequest implements RestModel {
```
plyhun, line 14: Can't we reuse the existing BinaryFieldTransformRequest, at it looks like identical?
Need also your opinions on this @bernhardriegler , @yrucrem .

rhoxhaj: Actually I agree with you @plyhun . When i started working on it, I planned there would be some change in the functionality, which didn't happen. I would opt for the usage of BinaryFieldTransformRequest.

bernhardriegler: @plyhun good suggestion - @rhoxhaj please go ahead and use the existing BinaryFieldTransformRequest

mdm/common/src/main/java/com/gentics/mesh/core/image/ImageManipulator.java
```diff
@@ -47,38 +47,36 @@
     Single<File> handleS3Resize(String bucketName, String s3ObjectKey, String filename, ImageManipulationParameters parameters);
```
plyhun, line 47: Here it looks like only the indentation changed, so at least this file can be reverted.

plyhun: UPD: caught the old commit, all ok in the end.

plyhun:
Looks ok to me.

PR #1193, opened 2021-04-23
Feature/ms3 13 metadata extraction

rhoxhaj:
## Abstract
This endpoint will do: 
- Mesh will download the binary
- extract and store metadata in the same way like binary data do.

Any suggestion regarding code refactoring to use the same binary processors is welcomed.

INFO: There was a problem with the rebase. I will squash the commits before merging.


core/src/main/java/com/gentics/mesh/core/s3binary/AbstractS3BinaryProcessor.java
```diff
@@ -0,0 +1,9 @@
+package com.gentics.mesh.core.s3binary;
+
+
+/**
+ * @see BinaryDataProcessor
+ */
+public abstract class AbstractS3BinaryProcessor implements S3BinaryDataProcessor {
+// TODO empty abstract class is not needed. Remove this
```
plyhun, line 8: Dont forget about this TODO please :smiley:

plyhun (changes requested):

core/src/main/java/com/gentics/mesh/core/endpoint/node/S3BinaryMetadataExtractionHandlerImpl.java
```diff
@@ -0,0 +1,332 @@
...
+									return fileName;
+								}
+
+								@Override
+								public String contentType() {
+									return mimeTypeForFilename;
+								}
+
+								@Override
+								public String contentTransferEncoding() {
```
plyhun, line 163: Are these two nulls below correctly processed? 
If null is not expected there as a special case, I would place some existing (if possible) or default values there. Charset is usually UTF-8. 
If it is not possible to get these values at the moment, I will test this part explicitly during the final merge.

plyhun (commented):
All look good, except the part with two dangling NULLs.

core/src/main/java/com/gentics/mesh/core/endpoint/node/S3BinaryMetadataExtractionHandlerImpl.java
rhoxhaj: I fixed it for the charset, but I cannot find any default value for the contentTransferEncoding. In any case I am not using these values at all.

PR #1197, opened 2021-04-30
Feature/ms 15 link resolving

rhoxhaj:
## Abstract
Extension of the link resolving for the S3 binary fields. In case when mesh configuration is set to "linkResolver:s3", the link resolver should resolve all S3 webroot paths. 
For a more detailed description: https://jira.gentics.com/browse/MS3-15


plyhun (approved):

services/aws-s3-storage/src/main/java/com/gentics/mesh/storage/s3/S3BinaryStorageImpl.java
```diff
@@ -142,7 +142,6 @@ public void createBucket(String bucketName) {
 		}
 
 		S3RestResponse s3RestResponse = new S3RestResponse(presignedRequest.url().toString(), presignedRequest.httpRequest().method().toString(), presignedRequest.signedHeaders());
-		s3RestResponse.setVersion(nodeVersion);
```
plyhun: I'm not yet sure if we may omit setting the node version explicitly. Will try keeping it in mind until the tests.

PR #1201, opened 2021-05-14
Make plugin startup aware of DB status + restart the init.

plyhun:
## Abstract

Mesh plugins are allowed to change the Mesh data over the MeshRestClient during the initialization. In the case of distributed database the DB availability has to be concerned during the plugin initialization. This MR applies a distributed DB topology lock listener, which triggers the currently initializing plugins to shutdown and restart once the DB is back again.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
@@ -587,13 +587,7 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 			PluginState pluginState = pluginWrapper.getPluginState();
 			if ((PluginState.DISABLED != pluginState) && (PluginState.STARTED != pluginState)) {
 				try {
-					log.info("Start plugin '{}'", getPluginLabel(pluginWrapper.getDescriptor()));
-
 					Plugin plugin = pluginWrapper.getPlugin();
-
-					//if (plugin.isUsingRestClient()) {
-						database.blockingTopologyLockCheck();
-					//}
 
 					plugin.start();
 					// Set state for PF4J
@@ -601,8 +595,21 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 					if (plugin instanceof MeshPlugin) {
 						// Set status for Mesh
 						MeshPlugin meshPlugin = (MeshPlugin) plugin;
-						setStatus(meshPlugin.id(), PluginStatus.STARTED);
+
+						//if (plugin.isUsingRestClient()) {
+							database.blockingTopologyLockCheck();
+						//}
+
 						pluginRegistry.preRegister(meshPlugin);
+
+						// If the plugin has failed to init once
+						if (getStatus(meshPlugin.id()) == PluginStatus.FAILED && source != failedPlugins) {
+							pluginWrapper.setPluginState(PluginState.RESOLVED);
+							setStatus(meshPlugin.id(), PluginStatus.LOADED);
+							throw new IllegalStateException("Plugin " + meshPlugin.id() + " failed to initialize once!");
+						} else {
+							setStatus(meshPlugin.id(), PluginStatus.STARTED);
+						}
 					}
 					startedPlugins.add(pluginWrapper);
 					
--- a/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
@@ -117,7 +117,6 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 
 	protected void delayedInitialize() {
 		super.initialize();
-		registerEventListeners();
 	}
 
 	protected void registerEventListeners() {
@@ -145,6 +144,7 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 					}
 				}
 			});
+			log.info("Distributed DB status event listener registered.");
 		}
 	}
 
@@ -165,6 +165,11 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 				startingPlugins.clear();
 			}
 		}
+	}
+	
+	@Override
+	public void init() {
+		registerEventListeners();
 	}
 
 	@Override
--- a/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
@@ -575,12 +575,7 @@ public class MeshPluginManagerImpl extends AbstractPluginManager implements Mesh
 
 			// 4. Register plugin
 			try {
-				PluginWrapper wrapper = getPlugin(id);
-				Plugin plugin = wrapper.getPlugin();
-				if (plugin instanceof MeshPlugin) {
-					MeshPlugin meshPlugin = (MeshPlugin) plugin;
-					pluginRegistry.preRegister(meshPlugin);
-				}
+				registerMeshPlugin(id);
 			} catch (Throwable e) {
 				log.error("Plugin registration failed with error", e);
 				try {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -588,4 +574,55 @@ public PluginResponse toResponse(MeshPlugin plugin) {
...
+		
+		for (PluginWrapper pluginWrapper : sourceCopy) {
+			PluginState pluginState = pluginWrapper.getPluginState();
+			if ((PluginState.DISABLED != pluginState) && (PluginState.STARTED != pluginState)) {
+				try {
+					log.info("Start plugin '{}'", getPluginLabel(pluginWrapper.getDescriptor()));
+
+					Plugin plugin = pluginWrapper.getPlugin();
+
+					//if (plugin.isUsingRestClient()) {
```
plyhun, line 594: I would really like to limit the application of the topolock check to the plugins, that do use REST API (i.e. DB access), although I see no way to check this before the plugin is initialized. Say, if the plugin requests MeshRestClient module, then we know it to potentially change the DB, but this is done on the plugin init the earliest, but that's too late for us. Question is open though.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -588,4 +574,55 @@ public PluginResponse toResponse(MeshPlugin plugin) {
...
+						source.remove(pluginWrapper);
+					}
+					
+					firePluginStateEvent(new PluginStateEvent(this, pluginWrapper, pluginState));
+				} catch (Throwable e) {
+					if (source == failedPlugins) {
+						log.error("Error while starting plugin {" + pluginWrapper.getPluginId() + "}: " + e.getMessage(), e);
+					} else {
+						log.warn("First error while starting plugin {" + pluginWrapper.getPluginId() + "}: " + e.getMessage() + ", giving a retry...", e);
+						failedPlugins.add(pluginWrapper);
```
plyhun, line 619: This approach to reinit failed plugins is IMO too basic, so if the reinit really helps and does not harm the overall Mesh startup process, I would make a more sophisticated version of it, including configs, timeouts and killswitch.

databases/orientdb/src/main/java/com/gentics/mesh/graphdb/cluster/TopologyEventBridge.java
```diff
@@ -113,14 +113,16 @@ public void onDatabaseChangeStatus(String nodeName, String iDatabaseName, DB_STA
 		}
 		databaseStatusMap.put(nodeName, iNewStatus);
 		log.info("Node {" + nodeName + "} Database {" + iDatabaseName + "} changed status {" + iNewStatus.name() + "}");
+		boolean isMe = "storage".equals(iDatabaseName) && nodeName.equals(manager.getNodeName());
```
bernhardriegler, line 116: `isMe` will be true if the current node in a cluster is the manager, right?
I suggest renaming it to `isManager`

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -149,20 +221,69 @@ public void startPlugins() {
...
+			log.debug("Already started plugin '{}'", getPluginLabel(pluginDescriptor));
+			return PluginState.STARTED;
+		}
+
+		if (!resolvedPlugins.contains(pluginWrapper)) {
+			log.warn("Cannot start an unresolved plugin '{}'", getPluginLabel(pluginDescriptor));
+			return pluginState;
+		}
+
+		// TODO Are we allowed to enable plugins by ourselves?
```
bernhardriegler, line 258: What is the alternative?
Why should a plugin not be enabled once we reach this point?

databases/orientdb/src/main/java/com/gentics/mesh/graphdb/cluster/TopologyEventBridge.java
plyhun: `manager.getNodeName()` returns the name of a *current* node:
```
	/**
	 * Determine the OrientDB Node name.
	 * 
	 * @return
	 */
	public String getNodeName() {
		StringBuilder nameBuilder = new StringBuilder();
		String nodeName = options.getNodeName();
		nameBuilder.append(nodeName);

		// Sanitize the name
		String name = nameBuilder.toString();
		name = name.replaceAll(" ", "_");
		name = name.replaceAll("\\.", "-");
		return name;
	}
```
So whenever the manager marks the current node as ready, the initial Mesh node starting process is unlocked.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
plyhun: This is the question. In the PF4J's code itself there are cases when the DISABLED status is set during the startup (i.e. if the plugin fails the validation). The plugin may be in theory disabled externally from the config.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -79,30 +90,92 @@
...
+	private final Lazy<Vertx> vertx;
+
 	private final PluginFactory pluginFactory;
 
 	private final MeshOptions options;
 
+	private final List<PluginWrapper> restartingPlugins = Collections.synchronizedList(new ArrayList<>(0));
+
+	private final List<PluginWrapper> startingPlugins = Collections.synchronizedList(new ArrayList<>(1));
+
```
plyhun, line 102: Those lists are synced because of a potential race conditions.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -79,30 +90,92 @@
...
+						handleDatabaseAvailability(true);
+						break;
+					}
+				}
+			});
+			log.info("Distributed DB status event listener registered.");
+		}
+	}
+
+	protected synchronized void handleDatabaseAvailability(boolean available) {
```
plyhun, line 151: This `synchronized` ensures that we don't mess with the multiple ONLINE state changes.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -79,30 +90,92 @@
...
+		if (available) {
+			startPluginsFrom(restartingPlugins);
+		} else {
+			synchronized (startingPlugins) {
```
plyhun, line 155: This `synchronized` ensures that we don't mess with non-ONLINE state changes and the initial plugin initialization. Those should never be called from each other, to avoid deadlocks.

core/src/main/java/com/gentics/mesh/plugin/manager/MeshPluginManagerImpl.java
```diff
@@ -149,20 +226,69 @@ public void startPlugins() {
...
+		if (pluginStorage == restartingPlugins) {
+			pluginStorage.clear();
+		}
+	}
+
+	/**
+	 * Start the specified plugin and its dependencies.
+	 */
+	@Override
+	public PluginState startPlugin(String pluginId) {
```
plyhun, line 245: This method is copied from `super` impl, with the modification of the *Mesh* plugin status notification - it should be performed only after the plugin is initialized, as we now have the separate state for initialization..

core/src/test/plugins/failing-first/src/main/java/com/gentics/mesh/plugin/FailingFirstPlugin.java
```diff
@@ -0,0 +1,36 @@
...
+	@Override
+	public Completable initialize() {
+		return Completable.defer(() -> {
+			try {
+				log.info("Failing starts up!");
+				environment().vertx().fileSystem().readFileBlocking(FILENAME);
+				return Completable.complete();
+			} catch (Exception e) {
+				log.error("Failing failed!", e);
+				Thread.sleep(1000 * 60 * 5);
```
plyhun, line 30: Hardcoded 5 minutes of the time-consuming high load task, when starting for the first time.

plyhun:
The functionality has been used and improved in https://github.com/gentics/mesh/pull/1205 . Closing this.

PR #1202, opened 2021-05-17
A routing strategy on the topology lock changes

plyhun:
## Abstract

Introduce the handle to control over the strategy of incoming request processing. when the (distributed) storage is not available for whatever reason.

1. Default behavior: all the requests pass and wait for the DB is available again for the defined timeout, as before.
2. Drop the DB writing requests (create/update/delete) with HTTP 503, when the DB is not available.

### General

### On API Changes


plyhun:
Reimplemented and merged as part of https://github.com/gentics/mesh/pull/1206 .  Closing.

PR #1203, opened 2021-05-18
SUP-11433: Log transaction retries better

plyhun:
## Abstract

Trace level of logging is useless for getting the reasons and details of the DB transaction retries. 

### General

### On API Changes


npomaroli:
This solution will probably spam the log:
It might be a quite normal situation, that OrientDB throws a ONeedRetryException (indicating that retrying will solve the problem), I do not think, that we should log this as an error every time.
Also when we hit the retry limit (which is 10 by default), 10 exceptions will have been logged.

I think, a better solution would be to keep and re-throw the last exception thrown by OrientDB, instead of throwing the RuntimeException at the end of this method. Or at least use that exception as "cause" for the RuntimeException.

PR #1214, opened 2021-06-11
SUP-11537/bugfix/fix permission issue in S3

rhoxhaj:
## Abstract
This is a bugfix for this bug: Caused by: graphql.AssertException: type S3BinaryField not found in schema
The bug is caused by not handling the field in GraphQL.
https://jira.gentics.com/browse/SUP-11537


plyhun (approved):

plyhun:
All good. Please make a port to hotfix-1.6.x as well.

PR #1222, opened 2021-07-23
Add BulkActionContext.process(force) to purge nodes process

kdima001:
When deleting old versions of nodes, files were not deleted from the disk. It seems that the reason is that there is no call to BulkActionContext.process after purging nodes.
Add BulkActionContext to purgeVersions method, push it to purgeVersion method, and finally call force process from BulkActionContext.

### General

### On API Changes


plyhun:
Hi @kdima001 ,
Huge thanks for the contribution!

Have couple of questions in addition:
1. Would you care of adding a dedicated unit test? If no - no problem, we will add it ourselves, it will just happen a bit later (having quite a workload currently).
2. Wouldn't you mind if we mention you name in the corresponding changelog line?

Thanks in advance!

kdima001:
Hi @plyhun.

1. I couldn't figure out how the cleanup test should work. I assume that you need to create multiple versions of the node (1.0, 1.1, 1.2, 2.0), call the cleanup and make sure that for versions 1.1, 1.2, there was an accumulation of deletion events in the BulkActionContext text and the process method was called. I think I should add the test to the PurgeHandler test of the project version. But I didn't find a similar test that could be used as an example. Perhaps I could write a test if you explained to me how it should work.

2. I don't mind

PR #1226, opened 2021-08-03
CL-697: Get rid of RootVertex concept from the common code 

plyhun:
... (everywhere except GraphFieldContainers)

## Abstract

Actions done here:
1. Introduce a DAO extension interface, marking the DAO as having a root-leaves connection from an external root entity -> RootDao<R, L>. Assign RootDao to the DAOs that have root relations (i.e. TagFamily -> Tag, Project -> Branch).
2. Get rid of (almost*) all *Root references in the BootsrapInitializer.
3. Replace (almost*) all the references to the *Root with the corresponding *Dao/*RootDao calls.
4. Rename DaoWrapper into Dao, for convenience. 
5. Turn around interface depencency `Dao extends DaoGlobal` into `DaoGlobal extends Dao`, as not all the DAOs allow global lookup (Node, Branch etc). This makes the implementors more consistent ([example](https://github.com/gentics/mesh/pull/1226/files#diff-e064a7b2396e62ffcadc83234de877d883f29ceade5199ec9bbc48202c303a3cL95)).
6. Suffix the methods of DaoGlobal for convenient recognizing.
7. (Optional / needs confirmation) Suffix the RootDao methods for convenient recognizing (cc @npomaroli do you agree on this? )

*where possible
### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/core/schema/MicroschemaEndpointTest.java
```diff
@@ -318,7 +318,7 @@ public void testUpdateWithBogusUuid() throws GenericRestException, Exception {
 
 			call(() -> client().updateMicroschema("bogus", request), NOT_FOUND, "object_not_found_for_uuid", "bogus");
 
-			Microschema reloaded = boot().microschemaContainerRoot().findByUuid(microschema.getUuid());
+			Microschema reloaded = boot().meshRoot().getMicroschemaContainerRoot().findByUuid(microschema.getUuid());
```
npomaroli, line 321: Can be rewritten like
HibMicroschema reloaded = boot().microschemaDao().findByUuidGlobal(microschema.getUuid());
(all occurrances)

mdm/api/src/main/java/com/gentics/mesh/core/data/dao/DaoGlobal.java
```diff
@@ -1,11 +1,14 @@
...
+import com.gentics.mesh.core.data.HibBaseElement;
+import com.gentics.mesh.core.result.Result;
+
 /**
  * Interface for DAO's which provide methods which allow installation wide queries on elements. The provided methods should not be scoped to a project, branch.
  * 
  * @param <T>
  */
-public interface DaoGlobal<T> {
+public interface DaoGlobal<T extends HibBaseElement> extends Dao<T> {
```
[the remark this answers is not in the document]
plyhun: Fair enough. I will check whether it is possible to remove the prefixes without interfering other functionality.

tests/tests-core/src/main/java/com/gentics/mesh/core/schema/MicroschemaEndpointTest.java
plyhun: IIRC I did this in the later pull requests... let me check.

plyhun: I did it for limited number of cases, to avoid pull requests overload. Let's see how much will it take to replace all the common code occurrences...

plyhun: Replaced where possible. The missing test cases will be anyway reviewed during the Enterprise project development.

PR #1229, opened 2021-08-18
Transaction generalizing

plyhun:
## Abstract

Tx entity is widely used in the common code, so has to be generalized into the MDM API as well.

Actions performed:
1. Tx split into general Tx and GraphDBTx. Tx went into mdm-api.
2. *DAOAction* classes went into mdm-api.
3. BaseTransaction class split into BaseTransaction and GraphDBBaseTransaction. BaseTransaction went into mdm-api.
4. DAOCollection now references *DAOs, not *DAOWrappers. DAOCollection went into mdm-api. 
5. All the references to DAOWrappers replaced with ones to DAOs *where possible*. Otherwise a classcast is used, which will be removed after the field containers are split.

### General

### On API Changes


mdm/api/src/main/java/com/gentics/mesh/core/db/BaseTransaction.java
```diff
@@ -0,0 +1,38 @@
+package com.gentics.mesh.core.db;
+
+public interface BaseTransaction extends AutoCloseable {
```
npomaroli, line 3: Javadoc

npomaroli (changes requested):

PR #1239, opened 2021-09-21
Add liveness probe by regularly touching a file.

npomaroli:
Change readiness probe to check status of local storage

### General

### On API Changes


plyhun:
@npomaroli 
AFAIK one cannot simply unit-test this functionality. What are the cases, that you have already tested manually?

npomaroli:
@plyhun I did manual tests in my local kubernetes with our mesh helm chart:
I changed the liveness probe in the helm chart to execute the script live.sh and reduced the initialDelaySeconds to 1
I verified that the started pod was "live" for kubernetes right after it was started and as long as the file was regularly touched
I verified (by manually deleting the file in the pod) that the pod was reported "unhealthy" and eventually got restarted by kubernetes
I also did a test for handling the OutOfMemory error by making a plugin which would occupy but never release a given amount of memory when requested. After requesting too much memory (OutOfMemory occurred), the liveness changed to "false".
I manually tested that changing the status of the local OrientDB "storage" to something different than "ONLINE" changes the readiness (by executing an OrientDB backup).

plyhun:
Ok. I'll grab this branch and run couple more non-cluster sanity checks to make sure there is no dependency to your local setup.

plyhun (approved):

PR #1254, opened 2021-10-13
WIP: JWT Options

deckdom:
Added JWT options to mesh config for more control of how tokens are generated and verified


api/src/main/java/com/gentics/mesh/etc/config/AuthenticationOptions.java
```diff
@@ -25,8 +25,12 @@
 public class AuthenticationOptions implements Option {
 
 	public static final String DEFAULT_ALGORITHM = "HS256";
+	public static final String DEFAULT_ISSUER = "Gentics Mesh";
```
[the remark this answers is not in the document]
plyhun: Since we in Mesh do not tell apart the client token and API token, this is also a breaking change regarding API token usage, which now awaits for the issuer+audience in payload during the API call. I would default both new fields to null.

common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
```diff
@@ -246,7 +247,8 @@ public String generateAPIToken(com.gentics.mesh.core.data.User user, String toke
 		JsonObject tokenData = new JsonObject()
 			.put(USERID_FIELD_NAME, user.getUuid())
 			.put(API_KEY_TOKEN_CODE_FIELD_NAME, tokenCode);
-		JWTOptions jwtOptions = new JWTOptions().setAlgorithm(options.getAlgorithm());
+		JWTOptions jwtOptions = new JWTOptions().setAlgorithm(options.getAlgorithm())
```
plyhun, line 250: Audience and issuer values are mandatory for Vert.x JWT auth implementation, so we have to
* either consider the Mesh settings, while authenticating the token (which may be the breaking change for the existing API tokens of the clients). This option is implemented below.
* or ignore them during the authentication process. Which makes the whole pull request kinda useless.

common/src/main/java/com/gentics/mesh/util/JWTUtil.java
```diff
@@ -0,0 +1,18 @@
+package com.gentics.mesh.util;
+
+import com.gentics.mesh.etc.config.AuthenticationOptions;
+
+import io.vertx.ext.jwt.JWTOptions;
+
+public class JWTUtil {
```
plyhun, line 7: Javadoc please.

core/src/test/java/com/gentics/mesh/core/user/JWTConfigurationTest.java
```diff
@@ -0,0 +1,88 @@
...
+			}}
+		});
+	}
+
+	@Override
+	public MeshTestContext getTestContext() {
+		return testContext.setOptionChanger(optionChanger);
+	}
+
+	@Test
```
plyhun, line 56: Need also an unhappy case (bad audience / bad issuer etc).

plyhun: Need also a leeway configuration test. Since the leeway setting is set explicitly on the Mesh start, it is not applied with OptionsChanger, and requires the separate test file.

api/src/main/java/com/gentics/mesh/etc/config/AuthenticationOptions.java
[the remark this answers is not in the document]
deckdom: I tested this scenario, and it is not breaking.
The JWT Authenticator does not automatically verify the received `Issuer` or `Audience` fields.
It's also stated in the JWT reference that these fields are optional and should be checked on an Application Layer instead.

I also think the information of `Issuer` and `Audience` is not really relevant for Mesh itself, but for clients/applications using Mesh.
The only place where this would be relevant imo would be in OAuth2 handlers, where we validate the token from another service (i.E. Keycloak).

PR #1260, opened 2021-10-27
Update to orientdb 3.1.14

udovidio:
## Abstract
Update to OrientDB version 3.1.14

### General

### On API Changes


npomaroli (changes requested):
Checks with old OrientDB storages looks good.
Please add a changelog entry.

npomaroli (approved):

PR #1262, opened 2021-11-03
LLDB DAO re-implemented (incomplete)

plyhun:
## Abstract

Get the low level DB manipulation methods back from Tx into DAOs, into the new Developer API layer ( `Persisting*Dao` ).

### General

### On API Changes


mdm/api/src/main/java/com/gentics/mesh/core/data/dao/ProjectDao.java
```diff
@@ -89,8 +86,8 @@ HibProject create(String name, String hostname, Boolean ssl, String pathPrefix,
 	ProjectSchemaEventModel onSchemaAssignEvent(HibProject project, HibSchema schema, Assignment assignment);
 
 	/**
-	 * Create a microschema assign event for the given input values.
-	 *
+	 * Create a schema assign event for the given input values.
```
npomaroli, line 89: It's about microschema here.

mdm/api/src/main/java/com/gentics/mesh/core/data/dao/SchemaDao.java
```diff
@@ -330,135 +116,31 @@ default void addSchema(HibSchema schemaContainer, HibProject project, HibUser us
...
-			.map(branch -> branch.onSchemaAssignEvent(version, UNASSIGNED, null));
-	}
+	Result<? extends HibNode> findNodes(HibSchemaVersion version, String uuid, HibUser user, ContainerType type);
+
+	/**
+	 * Add the schema to the db.
+	 * 
+	 * @param schema
+	 */
+	void addSchema(HibSchema schema);
```
npomaroli, line 145: Isn't this a low-level method and should be moved to PersistingSchemaDao?
Outside of PersistingSchemaDao, it is used by tests only.

mdm/common/src/main/java/com/gentics/mesh/core/db/CommonTx.java
```diff
@@ -0,0 +1,73 @@
...
+	@Override
+	PersistingBranchDao branchDao();
+
+	@Override
+	PersistingMicroschemaDao microschemaDao();
+
+	@Override
+	PersistingSchemaDao schemaDao();
+
+	BinaryDao binaryDao();
```
npomaroli, line 68: Add @Override

mdm/common/src/main/java/com/gentics/mesh/core/db/CommonTx.java
```diff
@@ -0,0 +1,73 @@
...
+
+	NodeDao nodeDao();
```
npomaroli, line 70: Add @Override

mdm/common/src/main/java/com/gentics/mesh/core/db/CommonTx.java
```diff
@@ -0,0 +1,73 @@
...
+
+	ContentDao contentDao();
```
npomaroli, line 72: Add @Override

npomaroli (changes requested):

mdm/api/src/main/java/com/gentics/mesh/core/data/dao/SchemaDao.java
plyhun: This method violates the DAO API strategy, making the duplicated things of the LLDB, and has been removed.

mdm/common/src/main/java/com/gentics/mesh/core/db/CommonTx.java
plyhun: This and the unchanged DAO accessors below have been removed because of being unneeded, as of now.

PR #1270, opened 2021-11-11
Permissions rewrite

plyhun:
## Abstract

The permission system should get rid of the deprecated code, making it easier to adapt to the 3rd party server implementations.

### Actions

* PermissionProperties(Impl) removed. The permissions management is performed by RoleDao.
* Permissions management is removed from the entities, since it duplicates the RoleDao functions.
* HasPermissions class removed, since it holds neither functionality nor properties anymore.
* New CommonTxData layer introduced, to prevent moving router related classes into API (Router classes are now accessed from the common BL layer, instead of being injected into DAOs). 
* RoleDao adapted accordingly.
* Unit tests adapted accordingly.
* Adhoc DAO fixes: 
  -- assign/unassign ContainerDao generic methods instead of add/remove + Schema/Microschema (i.e. just rename)
  -- found ProjectDao business logic moved into API
  -- helper `onRootDeleted()` method to deal with abandoned graph root nodes/tables, on root entity being removed
  -- restrict admin user/group deletion, as per old TODO

### General

### On API Changes


elasticsearch/src/main/java/com/gentics/mesh/search/index/node/NodeContainerTransformer.java
```diff
@@ -131,15 +130,20 @@ private void addParentNodeInfo(JsonObject document, HibNode parentNode) {
...
 				roleUuids.add(role.getUuid());
 			}
+
+			// Also add the roles which would grant read on published nodes if the container is published.
+			if (type == PUBLISHED) {
+				for (HibRole role : Tx.get().roleDao().getRolesWithPerm(node, READ_PUBLISHED_PERM)) {
+					roleUuids.add(role.getUuid());
+				}
+			}
+		} catch (NullPointerException e) {
```
plyhun, line 144: A questionnable part. During the docu generation either `Tx.get()` or `roleDao()` may return null.

rest-model/src/main/java/com/gentics/mesh/core/rest/schema/impl/StringFieldSchemaImpl.java
```diff
@@ -44,8 +48,24 @@ public StringFieldSchema setAllowedValues(String... allowedValues) {
...
+			} else if (allowedValues instanceof JsonArray) {
+				setAllowedValues(((JsonArray) allowedValues).stream().map(Object::toString).toArray(size -> new String[size]));
+			}  else {
+				throw new IllegalStateException("Unsupported allowed value type: " + allowedValues.getClass().getCanonicalName());
+			}
+		} else {
+			// What a restriction removal request comes from the REST API,
+			// it gives null instead of an empty array, so the request 
+			// gets eventually lost. In this case we give the empty array back.
+			allowedValues = new String[0];
```
npomaroli, line 68: This does not make any sense, since allowedValues isn't used after that line

PR #1274, opened 2021-11-19
Mdm development node

udovidio:
## Abstract
Extract most node related logic to the `PersistingNodeDao`.
CI looks good, there are still a couple of cleanups to be done but in my opinion they are no blocker for a merge.

### General

* [x] Added abstract that describes the change
* [x] Added changelog entry to `/CHANGELOG.adoc`
* [x] Ensured that the change is covered by tests
* [x] Ensured that the change is documented in the docs

### On API Changes

* [x] Checked if the changes are breaking or not
* [x] Added GraphQL API if applicable
* [x] Added Elasticsearch mapping if applicable

PR #1275, opened 2021-11-22
Mesh core with no DB dependency.

plyhun:
## Abstract

Core currently contains a lot of low-level references to the OrientDB implementation. This has to be abstracted out.

Done:
 * Project hierarchy reorganized:
 from  `mdm-api / mdm-common / mdm-orientdb-wrapper / api / common / core / server`
 into `mdm-api / mdm-common / api / common / core / mdm-orientdb-wapper / server`
The subprojects, that depend on core, were cleaned up as well.
 * Everything, that inevitably depends on OrientDB, is moved from Core to OrientDB wrapper (e.g. entity impls, OrientDB specific components)
 * Code refactored to use generic/persisting DAOs 
 * Tests refactored to use generic/persisting DAOs or Lowlevel DB API for test entity initialization. Inrefactorable tests moved into orientdb-test-context project.
 * Being overloaded with dependencies, the TxData now contains the reference to the Mesh instance, where all the necessary dependencies can be found.
 * Jobs refactored to the clearer interface/impl structure, so the corresponding impl for Enterprise should be much easier.
 * The BranchDao refactoring has been started. This was the actual trigger for the current PR, as branch appeared ho have a lot of DB references in its BL.
 * The BL has been removed out of graph entities/roots where possible.
 
In addition, the OrientDBMeshOptions class appeared to live in the common API, so has to be popped off within the Node refactoring task https://github.com/gentics/mesh/pull/1274/commits/8b1cf4810e9b22c7ddefa27def413a403e2bc896 .


common/src/main/java/com/gentics/mesh/dagger/MeshComponent.java
```diff
@@ -64,6 +62,14 @@
 		 */
 		Builder searchProviderType(@Nullable SearchProviderType type);
 
+		/**
+		 * Inject the own instance of {@link SearchWaitUtil}.
+		 * 
+		 * @param type
+		 * @return
+		 */
+		Builder searchWaitUtilSupplier(@Nullable Supplier<SearchWaitUtil> swUtil);
```
plyhun, line 71: The original implementation of `SearchWaitUtil` overriding used a special set of Dagger modules, set in OrientDBMeshComponent, and overridden in MeshTestComponent. Since MeshTestComponent cannot extend OrientDBMeshComponent anymore, a new module overriding should be introduced. Here we allow a selected module (SearchWaitUtil) to be overridden with an optional supplier.

core/src/main/java/com/gentics/mesh/changelog/highlevel/HighLevelChangesList.java
```diff
@@ -18,15 +16,9 @@
 @Singleton
 public class HighLevelChangesList {
```
plyhun, line 18: The tems were moved into the OrientDB implementation.

core/src/main/java/com/gentics/mesh/core/data/job/BranchMigrationJob.java
```diff
@@ -99,54 +80,54 @@ private BranchMigrationContext prepareContext() {
...
 			throw e;
 		}
 	}
 
-	@Override
-	protected Completable processTask() {
-		BranchMigration handler = mesh().branchMigrationHandler();
-
+	default Completable processTask(Database db) {
+		BranchMigration handler = db.tx(tx -> {
```
plyhun, line 99: By the time the job is actually started, there is no transaction happening, so we have to initiate our own one.

core/src/main/java/com/gentics/mesh/core/data/job/VersionPurgeJob.java
```diff
@@ -0,0 +1,89 @@
...
+import com.gentics.mesh.core.project.maintenance.ProjectVersionPurgeHandler;
+import com.gentics.mesh.core.rest.MeshEvent;
+import com.gentics.mesh.core.rest.event.job.ProjectVersionPurgeEventModel;
+import com.gentics.mesh.core.rest.job.JobStatus;
+
+import io.reactivex.Completable;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
+public interface VersionPurgeJob extends JobCore {
```
plyhun, line 22: The impl moved into the orientdb-wrapper.

core/src/main/java/com/gentics/mesh/core/endpoint/admin/consistency/repair/NodeDeletionGraphFieldContainerFix.java
```diff
@@ -1,177 +0,0 @@
...
-import com.syncleus.ferma.FramedGraph;
-
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
-
-/**
- * This fix will create the missing {@link Node} for {@link NodeGraphFieldContainer}'s which were affected by the deletion bug which was fixed in version
- * 0.18.3. Due to this bug the {@link Node} was deleted leaving the {@link NodeGraphFieldContainer} dangling in the graph.
- */
-public class NodeDeletionGraphFieldContainerFix {
```
plyhun, line 33: Fully moved into the ContentDaoWrapperImpl.

databases/orientdb/src/main/java/com/gentics/mesh/graphdb/cluster/TopologyEventBridge.java
```diff
@@ -136,6 +136,7 @@ public void onDatabaseChangeStatus(String nodeName, String iDatabaseName, DB_STA
 			statusInfo.put("node", nodeName);
 			statusInfo.put("database", iDatabaseName);
 			statusInfo.put("status", iNewStatus.name());
+			statusInfo.put("online", iNewStatus == DB_STATUS.ONLINE);
```
plyhun, line 139: Here we get rid of manual parsing of the OrientDB node status, which happens in the Core below.

doc/pom.xml
```diff
@@ -33,7 +33,7 @@
 	<dependencies>
 		<dependency>
 			<groupId>com.gentics.mesh</groupId>
-			<artifactId>mesh-core</artifactId>
+			<artifactId>mesh-mdm-orientdb-wrapper</artifactId>
```
plyhun, line 36: Docs still *heavily* depend on OrientDB...

elasticsearch/src/main/java/com/gentics/mesh/search/index/node/NodeIndexHandlerImpl.java
```diff
@@ -826,18 +825,22 @@ private IndexInfo createDummyIndexInfo(SchemaModel schema, String language) {
...
 	 */
-	public String generateVersion(NodeGraphFieldContainer container, String branchUuid, ContainerType type) {
+	public String generateVersion(HibNodeFieldContainer container, String branchUuid, ContainerType type) {
 		return getTransformer().generateVersion(container, branchUuid, type);
 	}
 
 	@Override
 	public Function<String, HibNode> elementLoader() {
-		return (uuid) -> HibClassConverter.toGraph(db).index().findByUuid(Node.class, uuid);
+		// TODO override in OrientDB implementation to use index for performance, if required
```
plyhun, line 834: The OrientDB implementation has been readdressed here, due to the currently missing mechanism of dependency override:
https://github.com/gentics/mesh/pull/1277

mdm/orientdb-api/src/main/java/com/gentics/mesh/core/data/MeshCoreVertex.java
```diff
@@ -20,6 +20,7 @@
 	 * @param batch
 	 *            Batch to which entries will be added in order to update the search index.
 	 * @return true if the element was updated. Otherwise false
+	 * @deprecated move to DAO or rename
```
plyhun, line 23: We should restrict direct entity changes - everything has to use corresponding DAO.

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/root/impl/BranchRootImpl.java
```diff
@@ -0,0 +1,118 @@
+package com.gentics.mesh.core.data.root.impl;
```
plyhun, line 1: This file has been actually moved from Core. Github treated in as created anew somehow.

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/dagger/OrientDBMeshComponent.java
```diff
@@ -42,6 +45,9 @@
 		@BindsInstance
 		Builder searchProviderType(@Nullable SearchProviderType type);
 
+		@BindsInstance
```
plyhun, line 48: Also continued in https://github.com/gentics/mesh/pull/1277 .

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/impl/MeshFactoryImpl.java
```diff
@@ -4,6 +4,7 @@
 import com.gentics.mesh.MeshFactory;
```
plyhun, line 4: Enterprise does not use MeshFactory for now, so the impl has been fully moved into OrientDB.

npomaroli (changes requested):
There is a problem when starting ServerRunner with an existing DB. When invoking the changelog, there is a NPE while getting the ChangelogRoot, because there seems to be no current transaction.

common/src/main/java/com/gentics/mesh/util/SearchWaitUtilProvider.java
```diff
@@ -0,0 +1,12 @@
+package com.gentics.mesh.util;
+
+/**
+ * An external provider for {@link SearchWaitUtil}.
+ * 
+ * @author plyhun
+ *
+ */
+public interface SearchWaitUtilProvider {
```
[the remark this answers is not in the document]
plyhun: Not anymore. Removing.

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/ElementResolvingRootDao.java
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * An extension to {@link RootDao} with the possibilities to reference persisted leaf entities upon the root entity.
+ * 
+ * @author plyhun
+ *
+ * @param <R>
+ * @param <L>
+ */
+public interface ElementResolvingRootDao<R extends HibCoreElement<? extends RestModel>, L extends HibCoreElement<? extends RestModel>> extends RootDao<R, L> {
```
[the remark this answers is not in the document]
plyhun: There are implementations (like TagDaos) that implement RootDao but do not implement PersistingRootDao, as the ability to connect leaf entities to the root is optional. Though their leaves still have to be referenced through the root, so technically this element resolving implementation might have been settled in RootDao. But since we do not expose any default implementations in the API, I had to introduce this extra extension in the common code, to share the element resolving implementation between PersistingRootDao implementors and RootDao (not persisting) implementors.

plyhun:
Codereview fixes landed in https://github.com/gentics/mesh/pull/1274/commits/931f96ba52ab1ec820f204ef3ad042631c9272c4 .

PR #1292, opened 2022-01-05
Mdm development tag

udovidio:
## Abstract
Extract tag and tag family to persisting dao. Also extract methods purge and isAutoPurgeEnabled to content dao

### General

* [X] Added abstract that describes the change
* [X] Ensured that the change is documented in the docs

### On API Changes

* [X] Checked if the changes are breaking or not
* [X] Added GraphQL API if applicable
* [X] Added Elasticsearch mapping if applicable


mdm/orientdb-api/src/main/java/com/gentics/mesh/core/data/root/TagFamilyRoot.java
```diff
@@ -2,59 +2,19 @@
...
-	void addTagFamily(TagFamily tagFamily);
-
 	/**
 	 * Return the project which is the root element of this tagfamily.
 	 * 
 	 * @return
 	 */
 	Project getProject();
 
+	void addTagFamily(TagFamily tagFamily);
```
npomaroli, line 19: Add Javadoc

npomaroli (changes requested):

mdm/orientdb-api/src/main/java/com/gentics/mesh/core/data/root/TagFamilyRoot.java
udovidio: done

PR #1311, opened 2022-02-09
Mdm schema migration

udovidio:
## Abstract
This merge requests contain the following changes:

* rework crud handlers to only run within a single "mesh transaction". This is because mdm transaction implementation doesn't handle "nested transactions" commit, so we had a few instances in which an event was sent and the consumer was reading stale data (since the parent transaction was not committed already)
* rework schema migration to set the new schema version immediately (also needed for mdm)
* bug fixes (mostly about tests which are not behaving well in mdm)

### General

### On API Changes

### DIFF
--- a/mdm/api/src/main/java/com/gentics/mesh/core/db/TxEventAction.java
+++ b/mdm/api/src/main/java/com/gentics/mesh/core/db/TxEventAction.java
@@ -1,0 +1,10 @@
+package com.gentics.mesh.core.db;
+
+import com.gentics.mesh.event.EventQueueBatch;
+
+@FunctionalInterface
+public interface TxEventAction<T> {
+
+	T handle(EventQueueBatch batch, Tx tx) throws Exception;
+
+}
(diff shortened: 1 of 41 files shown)


mdm/api/src/main/java/com/gentics/mesh/core/db/TxEventAction.java
```diff
@@ -0,0 +1,10 @@
+package com.gentics.mesh.core.db;
+
+import com.gentics.mesh.event.EventQueueBatch;
+
+@FunctionalInterface
+public interface TxEventAction<T> {
```
npomaroli, line 6: Add Javadoc to interface and method

mdm/api/src/main/java/com/gentics/mesh/core/db/TxEventAction0.java
```diff
@@ -0,0 +1,10 @@
+package com.gentics.mesh.core.db;
+
+import com.gentics.mesh.event.EventQueueBatch;
+
+@FunctionalInterface
+public interface TxEventAction0 {
```
npomaroli, line 6: Add Javadoc to Interface and method

npomaroli (changes requested):

npomaroli (approved):

PR #1315, opened 2022-02-22
Fix possible NPE when deleting a project

npomaroli:
### General

### On API Changes

### DIFF
--- a/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingNodeDao.java
+++ b/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingNodeDao.java
@@ -1022,7 +1022,7 @@ public interface PersistingNodeDao extends NodeDao, PersistingRootDao<HibProject
 		}
 		ContentDao contentDao = Tx.get().contentDao();
 		// Delete all initial containers (which will delete all containers)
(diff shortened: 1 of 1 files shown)


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingNodeDao.java
```diff
@@ -1022,7 +1022,7 @@ default void delete(HibNode node, BulkActionContext bac, boolean ignoreChecks, b
 		}
 		ContentDao contentDao = Tx.get().contentDao();
 		// Delete all initial containers (which will delete all containers)
-		for (HibNodeFieldContainer container : contentDao.getFieldContainers(node, INITIAL)) {
+		for (HibNodeFieldContainer container : contentDao.getFieldContainers(node, INITIAL).list()) {
```
plyhun, line 1025: Does it really make a difference?
The only implementor of `Result`, `TraversalResult` calls the wrapped iterable in all of these cases, and if that iterable is null, it will crash anyway.

npomaroli: Yes it does, I think it's because we are deleting the containers and the edges to it in the for loop.
The iterator itself was not null, but after deleting the first container, it returned NULL as second element and this caused the NPE when trying to delete it.

PR #1317, opened 2022-02-22
Mdm migrations

udovidio:
## Abstract
Rework schema and microschema migrations to avoid cloning fields which are not part of the new schema

### General

### On API Changes


plyhun (commented):
Looks ok. I run some new tests locally.

npomaroli (approved):

PR #1333, opened 2022-03-10
Node test fixes

plyhun:
### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/core/node/NodeEndpointTest.java
```diff
@@ -370,7 +369,7 @@ public void testCreateDeleteCreateWithUuid() throws Exception {
 	@Test
 	@Override
 	public void testCreateWithDuplicateUuid() throws Exception {
-		String nodeUuid = tx(() -> project().getUuid());
+		String nodeUuid = tx(() -> project().getBaseNode().getUuid());
```
plyhun, line 372: Since we cannot prove that the UUID is globally unique, without enumerating all the existing data, we have to change the test to check the current entity only.

tests/tests-core/src/main/java/com/gentics/mesh/core/node/NodeEndpointTest.java
```diff
@@ -1382,24 +1381,19 @@ public void testReadNodeVersionForBranch() {
...
-			// Update the schema
-			SchemaVersionModel schema = node.getSchemaContainer().getLatestVersion().getSchema();
-			schema.setSegmentField(null);
-			node.getSchemaContainer().getLatestVersion().setSchema(schema);
-			mesh().serverSchemaStorage().clear();
-			tx.success();
-		}
+		String nodeUuid = tx(tx -> { return folder("news").getUuid(); });
+		String schemaUuid = tx(tx -> { return folder("news").getSchemaContainer().getUuid(); });
+		SchemaResponse schema = call(() -> client().findSchemaByUuid(schemaUuid));
```
plyhun, line 1386: Changing the schema directly is invalid. A migration should be used.

npomaroli (approved):

PR #1358, opened 2022-04-13
Mdm clustering

udovidio:
### General

### On API Changes


core/src/main/java/com/gentics/mesh/cli/AbstractBootstrapInitializer.java
```diff
@@ -944,7 +949,60 @@ protected void initLanguageSet(LanguageSet languageSet) {
...
+		}
+	}
+
+	/**
+	 * The timeout used when initializing a clustered vertx instance. If the instance is not available during the provided
+	 * timeout, and exception will be thrown
+	 * @return
+	 */
+	public int getClusteredVertxInitializationTimeoutInSeconds() {
+		return 10;
```
plyhun, line 996: How critical is to have this configurable?

plyhun (approved):
Looks fair from my side, since no new code has been added. Need an opinion of @npomaroli .

api/src/main/java/com/gentics/mesh/etc/config/ClusterOptions.java
```diff
@@ -93,6 +95,11 @@ public class ClusterOptions implements Option {
 	@EnvironmentVariable(name = MESH_CLUSTER_TOPOLOGY_CHANGE_READONLY_ENV, description = "Override the topology change readonly flag.")
 	private boolean topologyChangeReadOnly = DEFAULT_TOPOLOGY_CHANGE_READONLY;
 
+	@JsonProperty(required = false)
+	@JsonPropertyDescription("The path of the hazelcast configuration path")
+	@EnvironmentVariable(name = MESH_CLUSTER_HAZELCAST_CONFIG_PATH_ENV, description = "Override the default hazelcast configuration path")
+	private String hazelcastConfigPath = DEFAULT_HAZELCAST_CONFIG_PATH;
```
npomaroli, line 101: Please add a changelog entry for this change.

npomaroli (changes requested):

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/cli/OrientDBBootstrapInitializerImpl.java
```diff
@@ -273,7 +226,7 @@ protected void initCluster(MeshOptions options, PostProcessFlags flags, boolean
 		boolean active = false;
 		while (!active) {
 			log.info("Waiting for hazelcast to become active");
-			active = manager.getHazelcastInstance().getLifecycleService().isRunning();
+			active = ((HazelcastClusterManager) getClusterManager()).getHazelcastInstance().getLifecycleService().isRunning();
```
npomaroli, line 229: We should avoid this cast (I saw the same cast in the other bootstrap initializer implementation).
Maybe we should move method HazelcastInstance getHazelcast(); from interface OrientDBClusterManager to ClusterManager and use that here and in the other bootstrap initializer implementation to get the Hazelcast instance?

npomaroli (changes requested):

api/src/main/java/com/gentics/mesh/etc/config/ClusterOptions.java
udovidio: Now that I think about it I think this configuration option should be moved to mdm since it's not used by the orientdb version

npomaroli: Do we need this configurable anyways? I suppose the configuration path is set by OrientDB, if OrientDB is creating the Hazelcast instance. So maybe we should just set it to the same value in MDM?

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/cli/OrientDBBootstrapInitializerImpl.java
udovidio: I went with the generic ClusterManager implementation because we might use different implementations for different versions (e.g. ZookeperClusterManager), and I wanted to avoid adding the vertx-hazelcast dependency in the mdm-api.
But if that's not a problem, we can add a dependency to vertx-hazelcast and return the hazelcast implementation, or add a method returning the hazelcast instance as suggested. Both are fine from my point of view

npomaroli: Currently, there are no plans to use a different cluster manage, and if we do that sometime in the future, we need to change a lot more than just that, so I would go for the simplest and cleanest solution for now.

api/src/main/java/com/gentics/mesh/etc/config/ClusterOptions.java
udovidio: Yes, we could check the system property hazelcast.config, if not set, we hard code it to config/hazelcast.xml.
This way we allow user to override it if necessary

udovidio: I removed the configuration option

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/cli/OrientDBBootstrapInitializerImpl.java
udovidio: I went with exposing the HazelcastInstance in ClusterManager

core/src/main/java/com/gentics/mesh/cli/AbstractBootstrapInitializer.java
udovidio: I would avoid making it user-configurable for now, since it requires more mantainance work. Once we have the final version of mdm, we can evaluate whether we need more fine grained configuration from the user

npomaroli (approved):

PR #1361, opened 2022-04-15
highlevel changelog implementation for mdm mdm

udovidio:
### General

### On API Changes


plyhun (approved):

core/src/main/java/com/gentics/mesh/changelog/highlevel/HighLevelChangesList.java
```diff
@@ -1,40 +1,20 @@
...
-import com.gentics.mesh.changelog.highlevel.change.ExtractPlainText;
-import com.gentics.mesh.changelog.highlevel.change.SetAdminUserFlag;
 import com.gentics.mesh.core.data.changelog.HighLevelChange;
 
 /**
  * List of high level changes which can be applied once the low level changelog entries have been applied.
  */
 @Singleton
-public class HighLevelChangesList {
+public interface HighLevelChangesList {
```
npomaroli, line 13: Maybe it would be more flexible to make this class abstract and leave the (common) high level changes in here?
Context specific implementations (like OrientDBHighLevelChangesList) should just make sure, that they call super.getList() in their getList() implementation and add the DB specific changes to the common list.
Currently, 50% of the high-level changes we have would apply for both the OrientDB and the MDM version.

udovidio: If we will have common high-level changes, than yes, it makes sense. I will implement that change

udovidio: done

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/changelog/highlevel/OrientDBHighLevelChangesList.java
```diff
@@ -19,36 +16,30 @@
...
-			setAdminUserFlag
+		List<HighLevelChange> changeList = super.getList();
+
 		// ADD NEW CHANGES HERE!
 		// WARNING!
 		// Only add changes when absolutely needed. Try to avoid changelog entries since those would require a offline update.
 		// Instead it is recommended to write changes in a way so that those can be applied on the fly.
 		// WARNING!
-		);
+		changeList.add(restructureWebroot);
```
npomaroli, line 40: this will throw an UnsupportedOperationException, since the list was generated with Arrays.asList() in super.getList()

npomaroli (changes requested):

npomaroli (approved):

PR #1363, opened 2022-04-29
Refactoring to improve mdm performance

udovidio:
## Abstract
Refactored API code to allow mdm performance improvements in the following areas 
- graphql breadcrumbs call
- graphql nodes call
- node creation
### General

### On API Changes


common/src/main/java/com/gentics/mesh/router/ProjectsRouterImpl.java
```diff
@@ -68,7 +68,7 @@ public Router addProjectRouter(String name) throws InvalidNameException {
 			projectRouters.put(name, projectRouter);
 			log.info("Added project router {" + name + "}");
 
-			projectRouter.route().handler(ctx -> {
+			projectRouter.route().blockingHandler(ctx -> {
```
udovidio, line 71: during my load tests I found that the db could block for a long time here and block the event loop thread

npomaroli: You should use the call blockingHandler(handler, boolean) and set the boolean (ordered) to "false". Otherwise, such calls might get queued, even if enough worker pool threads are available.

PR #1367, opened 2022-05-03
Test leftovers

plyhun:
### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/core/field/AbstractFieldEndpointTest.java
```diff
@@ -78,9 +79,12 @@ protected NodeResponse updateNode(String fieldKey, Field field) {
 	 * @return
 	 */
 	protected NodeResponse updateNode(String fieldKey, Field field, boolean expandAll) {
-		tx(() -> prepareTypedSchema(schemaContainer("folder"), Optional.ofNullable(field).stream()
+		tx(tx -> {
+			prepareTypedSchema(schemaContainer("folder"), Optional.ofNullable(field).stream()
 				.map(TestHelper::fieldIntoSchema)
-				.map(schema -> schema.setName(fieldKey)).collect(Collectors.toList()), Optional.empty()));
+				.map(schema -> schema.setName(fieldKey)).collect(Collectors.toList()), Optional.empty()); 
+			tx.success();
```
npomaroli, line 86: This should not be necessary, because the method tx(handler) internally sets the success flag to the transaction, if no exception was thrown from within the handler

PR #1380, opened 2022-05-25
Fix various tests

npomaroli:
### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/test/context/AbstractMeshTest.java
```diff
@@ -101,7 +101,6 @@ public OkHttpClient httpClient() {
 
 	public String getJson(HibNode node) throws Exception {
 		InternalActionContext ac = mockActionContext("lang=en&version=draft");
-		ac.data().put(SharedKeys.PROJECT_CONTEXT_KEY, TestDataProvider.PROJECT_NAME);
```
plyhun, line 104: What is the purpose for this change?

npomaroli: The data should contain an instance of the project, not the name of the project stored with this key.
The call to mockActionContext() already puts the project into the data, but this is incorrectly overwritten here.

plyhun (approved):

PR #1388, opened 2022-06-07
Tests for SUP-13687

plyhun:
### General

### On API Changes

### DIFF
--- a/tests/tests-core/src/main/java/com/gentics/mesh/core/tag/TagNodeEndpointTest.java
+++ b/tests/tests-core/src/main/java/com/gentics/mesh/core/tag/TagNodeEndpointTest.java
@@ -47,25 +47,28 @@ public class TagNodeEndpointTest extends AbstractMeshTest {
 
 	@Test
 	public void testReadPublishedNodesForTag() {
+		String colorsUuid = tx(() -> tagFamily("colors").getUuid());
+		String redUuid = tx(() -> tag("red").getUuid());
+		
+		call(() -> client().takeNodeOffline(PROJECT_NAME, tx(() -> project().getBaseNode().getUuid()), new PublishParametersImpl().setRecursive(true)));
+		NodeListResponse nodeList = call(() -> client().findNodesForTag(PROJECT_NAME, colorsUuid, redUuid,
+			new VersioningParametersImpl().published()));
+		assertThat(nodeList.getData()).as("Published tagged nodes").isNotNull().isEmpty();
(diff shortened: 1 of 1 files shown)


mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
```diff
@@ -250,7 +250,9 @@ public Stream<Node> getChildrenStream(InternalActionContext ac, InternalPermissi
 		Tx tx = GraphDBTx.getGraphTx();
 		UserDao userDao = tx.userDao();
 		return toStream(getUnframedChildren(tx.getBranch(ac).getUuid()))
-			.filter(node -> userDao.hasPermissionForId(user, node.getId(), perm))
+			.filter(node -> perm != null
```
plyhun, line 253: Fallback to the old behavior when no perm parameter provided.

tests/tests-core/src/main/java/com/gentics/mesh/test/context/MeshTestContext.java
```diff
@@ -455,7 +455,9 @@ private void cleanupFolders() throws IOException {
 		for (File folder : tmpFolders) {
 			FileUtils.deleteDirectory(folder);
 		}
-		meshDagger.permissionCache().clear(false);
+		if (meshDagger != null && meshDagger.permissionCache() != null) {
```
plyhun, line 458: Cleanup fix.

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
npomaroli: This does not make any sense. The method is never called without perm from productive code, only from the testcase NodeTest.testGetChildrenStream(), so it is better to fix that testcase.

npomaroli (changes requested):

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/node/impl/NodeImpl.java
npomaroli: Testcase has now been fixed in dev branch.

npomaroli (approved):

PR #1389, opened 2022-06-08
OrientDB Direct Memory tuning.

plyhun:
## Abstract

Since OrientDB 3.1.18 a more aggressive RAM allocation strategy is implemented, leaving fewer memory for the rest of the host OS / container. A set of options entries introduced to control the allocation.

### General

### On API Changes


plyhun:
Fixed in the code for in-memory runtime occurrences.

PR #1399, opened 2022-06-27
If setup of mesh context fails, the test execution will fail immediately

npomaroli:
Only perform consistency checks on succeeded tests

## Abstract

If setup of the mesh test context fails, the whole test execution should fail immediately. If MeshTestContext is a TestWatcher, this cannot be achieved (TestWatcher implementations only observe the test execution, but do not modify it).

### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/test/context/MeshTestContext.java
```diff
@@ -187,18 +200,18 @@ private static OkHttpClient createTestClient() {
 			// Create a trust manager that does not validate certificate chains
 			final TrustManager[] trustAllCerts = new TrustManager[] {
 				new X509TrustManager() {
-					@Override
+				@Override
```
plyhun, line 203: Broken formatting.

plyhun (approved):

PR #1405, opened 2022-07-01
SUP-12299: JWT Options (1.6.x)

deckdom:
For 1.8 changes, see #1408

## Abstract

Additional options which will be passed to the JWT service which will be used for signing and validation.
Following Options are getting included:

* Leeway (Time in Seconds how much the expiary may be allowed)
* Issuer (Name of the Signer/Issuer of the Token)
* Audience (List of names which are valid targets of the Token)
* Ignore Expiration (Disable validation of expiration date entirely)

The fields Issuer and Audience are only meta fields and are not actually used in our Validation.
It is instead intended for Clients connecting to Mesh to further verify authenticity and/or check for targets.

See previous MR as well: https://github.com/gentics/mesh/pull/1254

### General

* [X] Added abstract that describes the change

### On API Changes

* [X] Added GraphQL API if applicable (none)
* [X] Added Elasticsearch mapping if applicable (none)


common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
```diff
@@ -44,28 +45,31 @@ public class MeshJWTAuthProvider implements AuthProvider, JWTAuth {
...
 	public MeshJWTAuthProvider(Vertx vertx, MeshOptions meshOptions, BCryptPasswordEncoder passwordEncoder, Database database, BootstrapInitializer boot) {
+		this.vertx = vertx;
 		this.meshOptions = meshOptions;
 		this.passwordEncoder = passwordEncoder;
 		this.db = database;
 		this.boot = boot;
+	}
+
+	public void initialize() {
+		if (this.jwtProvider != null) {
```
udovidio, line 70: can this really happen? I'd rather declare the provider as final and remove this check

LTS-CHANGELOG.adoc
```diff
@@ -20,6 +20,8 @@ All fixes and changes in LTS releases will be released the next minor release. C
 [[v1.6.30]]
 == 1.6.30 (TBD)
 
+icon:plus[] Auth: Added JWT options to configure token content (`issuer` - `"iss"`, `audience` - `"aud"`) and token verification (`leeway` and `ignoreExpiration`).
```
udovidio, line 23: You're a bit out of date with the origin branch, latest relased version is 1.6.34, so this should be 1.6.35

common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
deckdom: This initialization function is only needed for unit-testing.
The Provider is only getting initialized once the first request is received instead directly on startup.
In the tests this is required, as we need to replace the options once the server is already started.

udovidio (approved):

PR #1434, opened 2022-08-31
Implement missing S3 binary support for index engine, causing publish process to fail

plyhun:
## Abstract

Manipulating S3 fields with an index engine ends up in various errors within Mesh and its callers. This fixes the index engine support.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/endpoint/node/S3BinaryMetadataExtractionHandlerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/endpoint/node/S3BinaryMetadataExtractionHandlerImpl.java
@@ -73,7 +73,8 @@ public class S3BinaryMetadataExtractionHandlerImpl extends AbstractHandler {
 
 	@Inject
 	public S3BinaryMetadataExtractionHandlerImpl(Database db, S3BinaryStorage s3BinaryStorage, HandlerUtilities utils,
-												 Vertx rxVertx, MeshOptions options, S3Binaries s3binaries, S3BinaryProcessorRegistryImpl s3binaryProcessorRegistry) {
+			Vertx rxVertx, MeshOptions options, S3Binaries s3binaries,
+			S3BinaryProcessorRegistryImpl s3binaryProcessorRegistry) {
 		this.db = db;
 		this.vertx = rxVertx;
 		this.s3BinaryStorage = s3BinaryStorage;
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/gentics/mesh/core/endpoint/node/S3BinaryMetadataExtractionHandlerImpl.java
```diff
@@ -73,7 +73,8 @@ public class S3BinaryMetadataExtractionHandlerImpl extends AbstractHandler {
 
 	@Inject
 	public S3BinaryMetadataExtractionHandlerImpl(Database db, S3BinaryStorage s3BinaryStorage, HandlerUtilities utils,
-												 Vertx rxVertx, MeshOptions options, S3Binaries s3binaries, S3BinaryProcessorRegistryImpl s3binaryProcessorRegistry) {
```
plyhun, line 76: Formatting fixes in this file.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/s3binary/S3BinaryFieldEndpointTest.java
```diff
@@ -54,13 +63,26 @@ public class S3BinaryFieldEndpointTest extends AbstractFieldEndpointTest {
...
     @Before
     public void updateSchema() throws IOException {
-        try (Tx tx = tx()) {
-            S3BinaryFieldSchema s3BinaryFieldSchema = new S3BinaryFieldSchemaImpl();
-            s3BinaryFieldSchema.setName(FIELD_NAME);
-            s3BinaryFieldSchema.setLabel("Some label");
-            prepareTypedSchema(schemaContainer("content"), List.of(s3BinaryFieldSchema), Optional.empty());
-            tx.success();
-        }
+    	HibSchema schemaContainer = schemaContainer("content");
```
plyhun, line 66: Here we use REST API to update the schema, because it triggers the indexing engine to get updated with the new field data.

plyhun:
Not sure if we still support 1.6.

udovidio:
Yes, we do.
We must port it if the bug is affecting that version

plyhun:
I will then take care of cherrypicks onto all supported branches (1.6, 1.8).

udovidio (approved):

PR #1445, opened 2022-09-26
SUP-12782

udovidio:
Port of #1444 

### General

### On API Changes

### DIFF
--- a/LTS-CHANGELOG.adoc
+++ b/LTS-CHANGELOG.adoc
@@ -22,6 +22,9 @@ All fixes and changes in LTS releases will be released the next minor release. C
 
 icon:check[] Core: when an initial admin password is provided with an environment variable, its value won't be logged anymore.
 
+icon:check[] Core: when a binary file was manually deleted from the file system and the same file was uploaded again through the rest api, the binary
+upload endpoint was returning a success response even though the binary wasn't actually uploaded. The file will now be uploaded correctly.
+
 [[v1.8.10]]
 == 1.8.10 (22.09.2022)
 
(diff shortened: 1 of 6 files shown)


LTS-CHANGELOG.adoc
```diff
@@ -22,6 +22,9 @@ All fixes and changes in LTS releases will be released the next minor release. C
 
 icon:check[] Core: when an initial admin password is provided with an environment variable, its value won't be logged anymore.
 
+icon:check[] Core: when a binary file was manually deleted from the file system and the same file was uploaded again through the rest api, the binary
```
plyhun, line 25: This perhaps requires a single-line entry.

core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryUploadHandlerImpl.java
```diff
@@ -189,12 +190,21 @@ public void handleUpdateField(InternalActionContext ac, String nodeUuid, String
...
+					})
+					.runInNewTx();
+			HibBinary binary = binaryAndUuid.v1();
+			String binaryUuid = binaryAndUuid.v2();
 
 			// Create a new binary uuid if the data was not already stored
 			if (binary == null) {
 				ctx.setBinaryUuid(UUIDUtil.randomUUID());
 				ctx.setInvokeStore();
+			} else if (binaryUuid != null && fileNotExists(binaryUuid)) {
```
plyhun, line 205: Why can't we use UUID from the binary right here without a Tuple?

plyhun (changes requested):

core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryUploadHandlerImpl.java
udovidio: We are not in a transaction context, so binary.getUuid would through a null pointer exception. An alternative would be to open a separate transaction just to get the uuid, but it would bring some overhead

plyhun: Fair.

PR #1447, opened 2022-09-29
SUP-13112 do not check permission when users create their own token

udovidio:
### General

### On API Changes


udovidio:
After discussion IRL, we have decided that it still makes sense to check for user permission when creating a token, therefore this ticket can be closed

PR #1452, opened 2022-10-07
Dev gpu 311 permissions endpoints

npomaroli:
## Abstract

The current endpoints to change role permissions allow changing only for a single role. The new endpoints allow reading/granting/revoking for multiple (all) roles in a single request.

### General

### On API Changes


mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/dao/impl/RoleDaoWrapperImpl.java
```diff
@@ -72,6 +101,24 @@ public boolean revokeRolePermissions(HibRole role, HibBaseElement element, Inter
...
+	public boolean revokeRolePermissions(Set<HibRole> roles, HibBaseElement element, InternalPermission... permissions) {
+		MeshVertex vertex = (MeshVertex) element;
+
+		boolean permissionRevoked = false;
+		for (InternalPermission permission : permissions) {
+			Set<String> allowedRoles = getRoleUuidsForPerm(vertex, permission);
+			if (allowedRoles != null) {
+				for (HibRole role : roles) {
+					permissionRevoked = allowedRoles.remove(role.getUuid()) || permissionRevoked;
+					vertex.setRoleUuidForPerm(permission, allowedRoles);
```
npomaroli, line 114: Move this after the for loop

rest-model/src/main/java/com/gentics/mesh/core/rest/Examples.java
```diff
@@ -276,6 +276,13 @@ public static RoleReference roleRef() {
 		return ref;
 	}
 
+	public static RoleReference roleRef2() {
+		RoleReference ref = new RoleReference();
+		ref.setUuid(uuid4());
```
npomaroli, line 281: Use different UUID here

PR #1465, opened 2022-11-09
SUP-14400: Filter out indices to be cleaned on resync

plyhun:
## Abstract
When you have a schema, which has different settings per language: elasticsearch._meshLanguageOverride.[langCode], then Mesh will create a separate index for it.
i.E. `mesh-node-<project>-<branch>-<schemaVersion>-<draft/published>[-langCode][-microSchemaHash]`

The default/fallback index simply does not include the additional [-langCode].

The issue is, when you trigger a sync POST /search/sync.
This should verify all indices and update the documents accordingly (Adding missing ones, updating not up to date ones, and deleting documents which shouldn't be there).

However, this only seems to work on the default/fallback index.
The content of these Language Indices are dropped entirely and will then be re-populated.
This is not how the sync should work and causes the search for these languages to be unavailable/not properly useable until everything has been re-populated.

### General

### On API Changes

### DIFF
--- a/tests/tests-core/src/main/java/com/gentics/mesh/search/LanguageOverrideSearchTest.java
+++ b/tests/tests-core/src/main/java/com/gentics/mesh/search/LanguageOverrideSearchTest.java
@@ -2,6 +2,7 @@ package com.gentics.mesh.search;
 
 import static com.gentics.mesh.core.rest.common.ContainerType.DRAFT;
 import static com.gentics.mesh.core.rest.common.ContainerType.PUBLISHED;
+import static com.gentics.mesh.test.ClientHelper.call;
 import static com.gentics.mesh.test.TestDataProvider.PROJECT_NAME;
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -11,6 +12,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


tests/tests-core/src/main/java/com/gentics/mesh/search/LanguageOverrideSearchTest.java
```diff
@@ -182,6 +189,50 @@ public void testLanguageOverride() {
...
+			docs("zh", DRAFT, 1),
+			docs("ja", DRAFT, 1),
+			docs("ko", DRAFT, 1),
+			// italian and english contents use default settings
+			docs(microschemaHash.toLowerCase(), DRAFT, 3)
+		));
+
+		SyncEventHandler.invokeSyncCompletable(meshApi()).blockingAwait(30, TimeUnit.SECONDS);
+
+		assertDocumentCount(fromEntries(
```
deckdom, line 218: Not sure if this will work, as the content of the indices is cleared _during_ the sync.
In the test, you'd need always need to be "lucky" to get the timing right I think?

plyhun: This is how it works in all the tests around, just wrapped by `assertContentCount()`.
Since `invokeSyncCompletable()` is awaited for 30 seconds the most (which has to be enough for 1 page case), no race is expected.

plyhun: Also, the original usecase is reproducible with this test, if the fix is removed.

npomaroli:
As bugfix, this change must go to hotfix-1.8.x, so this PR has been replaced by #1466

PR #1470, opened 2022-11-28
SUP-14472: Two nodes with same path/segment field if binary is filled

plyhun:
## Abstract

Reproducible in the mesh-ui:

    Create a new Schema "test"
    with a field "id" type string and binary with name "binary"
    set segment field to id
    Create a new project "test" and link schema
    Create a new node with schema "test" and fill id with "1" -> save
    Create a new node and fill id with "1" again -> save -> ERROR because segment field already used
    Edit first created node and upload a binary in the binary field -> save
    Try again to create a new node with id "1" -> save -> WORKS
    => now you have two nodes with the same segment field

### General

### On API Changes

### DIFF
--- a/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingContentDao.java
+++ b/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingContentDao.java
@@ -235,10 +235,11 @@ public interface PersistingContentDao extends ContentDao {
 			// remove existing draft edge
 			if (draftEdge != null) {
 				removeEdge(draftEdge);
-				updateWebrootPathInfo(newContainer, branchUuid, "node_conflicting_segmentfield_update");
 			}
 			// create a new draft edge
 			createContainerEdge(node, newContainer, branchUuid, languageTag, DRAFT);
+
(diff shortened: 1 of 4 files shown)


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingContentDao.java
```diff
@@ -235,10 +235,11 @@ default void connectFieldContainer(HibNode node, HibNodeFieldContainer newContai
 			// remove existing draft edge
 			if (draftEdge != null) {
 				removeEdge(draftEdge);
-				updateWebrootPathInfo(newContainer, branchUuid, "node_conflicting_segmentfield_update");
 			}
 			// create a new draft edge
 			createContainerEdge(node, newContainer, branchUuid, languageTag, DRAFT);
+
```
plyhun, line 241: Because of `updateWebrootPathInfo` being called before creating a new edge, its changes never get indexed upon upload. The upload itself calls `updateWebrootPathInfo` only if the binary field being uploaded is the segment field.

npomaroli (approved):

PR #1473, opened 2022-12-06
SUP-14448: Mesh: LivenessCheck returns 200 even if EventbusCheck fails

plyhun:
## Abstract

If the cluster manager reports unresponsive pod on an own instance name, the instance should be considered "broken" and liveness check should fail so a restart can be triggered from an external System(Kubernetes, monitoring).

### General

### On API Changes

### DIFF
--- a/LTS-CHANGELOG.adoc
+++ b/LTS-CHANGELOG.adoc
@@ -20,7 +20,7 @@ All fixes and changes in LTS releases will be released the next minor release. C
 [[v1.6.40]]
 == 1.6.40 (TBD)
 
-icon:check[] Clustering: Liveness check now marks an instance as dead, if the cluster manager (Hazelcast) reports it as unresponsive, even if the non-cluster liveness check passes.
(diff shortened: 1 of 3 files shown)


LTS-CHANGELOG.adoc
```diff
@@ -17,6 +17,11 @@ include::content/docs/variables.adoc-include[]
 The LTS changelog lists releases which are only accessible via a commercial subscription.
 All fixes and changes in LTS releases will be released the next minor release. Changes from LTS 1.4.x will be included in release 1.5.0.
 
+[[v1.6.40]]
+== 1.6.40 (TBD)
+
+icon:check[] Clustering: Liveness check now marks an instance as dead, if the cluster manager (Hazelcast) reports it as unresponsive, even if the non-cluster liveness check passes.
```
npomaroli, line 23: The description is inacurate. Hazelcast is not directly involved in this check and we don't know whether the problem has something to do with it at all.
Actually, we are extending the EventBus Liveness check (mentioned in the changelog of version 1.6.28) to also set the liveness to false, if no ping from itself is received within the timeout over the cluster.

npomaroli (changes requested):

PR #1478, opened 2023-01-18
Bypass update webroot info for absent segment field value

plyhun:
## Abstract

Segment field should be checked for conflicts only if the segment field name is set.

### General

### On API Changes


plyhun:
Subject for 1.6.x+

PR #1479, opened 2023-01-18
Reset segment info, if segment field is null

plyhun:
## Abstract

If the segment field name is empty, the segment info is treated as outdated and is subject to reset.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/data/container/impl/NodeGraphFieldContainerImpl.java
+++ b/core/src/main/java/com/gentics/mesh/core/data/container/impl/NodeGraphFieldContainerImpl.java
@@ -390,8 +390,8 @@ public class NodeGraphFieldContainerImpl extends AbstractGraphFieldContainerImpl
 		String segment = node.getPathSegment(branchUuid, type, getLanguageTag());
 
 		// The webroot uniqueness will be checked by validating that the string [segmentValue-branchUuid-parentNodeUuid] is only listed once within the given
-		// specific index for (drafts or published nodes)
-		if (segment != null) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/gentics/mesh/core/data/container/impl/NodeGraphFieldContainerImpl.java
```diff
@@ -390,8 +390,8 @@ private boolean updateWebrootPathInfo(Node node, GraphFieldContainerEdge edge, S
 		String segment = node.getPathSegment(branchUuid, type, getLanguageTag());
 
 		// The webroot uniqueness will be checked by validating that the string [segmentValue-branchUuid-parentNodeUuid] is only listed once within the given
-		// specific index for (drafts or published nodes)
-		if (segment != null) {
+		// specific index for (drafts or published nodes)Segment field should also be set, otherwise the segment info is treated as outdated and is subject to reset.
+		if (segment != null && StringUtils.isNotBlank(segmentFieldName)) {
```
npomaroli, line 394: Could we make the test for the segmentFieldName earlier? Like it is done in the branch hotfix-1.9.x-webroot-bypass

plyhun: Then we are completely skipping the reset of `setSegmentInfo()`, which is not good for the case of segment field name removal. Or do you mean placing the `setSegmentInfo()` to the early return as well?

npomaroli: Ah, I see. We still need to go to the "else" branch, if there is no segment field name. That's ok then.

npomaroli (approved):

PR #1482, opened 2023-01-24
Chunked migration process

plyhun:
## Abstract

As of now the migration process tries to load all the data to process at once, which exhausts resources. We need to split the workload into chunks reliably.

### General

### On API Changes

### DIFF
--- a/api/src/main/java/com/gentics/mesh/etc/config/ContentConfig.java
+++ b/api/src/main/java/com/gentics/mesh/etc/config/ContentConfig.java
@@ -14,8 +14,10 @@ import com.gentics.mesh.etc.config.env.Option;
 public class ContentConfig implements Option {
 
 	public static final String MESH_CONTENT_AUTO_PURGE_ENV = "MESH_CONTENT_AUTO_PURGE";
+	public static final String MESH_GRAPH_BATCH_SIZE_ENV = "MESH_GRAPH_BATCH_SIZE";
 
-	private static final boolean DEFAULT_AUTO_PURGE = true;
+	public static final boolean DEFAULT_AUTO_PURGE = true;
+	public static final int DEFAULT_BATCH_SIZE = 0;
 
 	@JsonProperty(required = false)
 	@JsonPropertyDescription("Flag which controls the global setting for the auto purge mechanism. The setting can be overriden by the schema 'autoPurge' flag. Default: "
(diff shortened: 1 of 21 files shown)


api/src/main/java/com/gentics/mesh/etc/config/ContentConfig.java
```diff
@@ -14,19 +14,36 @@
 public class ContentConfig implements Option {
 
 	public static final String MESH_CONTENT_AUTO_PURGE_ENV = "MESH_CONTENT_AUTO_PURGE";
+	public static final String MESH_GRAPH_BATCH_SIZE_ENV = "MESH_GRAPH_BATCH_SIZE";
 
-	private static final boolean DEFAULT_AUTO_PURGE = true;
+	public static final boolean DEFAULT_AUTO_PURGE = true;
+	public static final int DEFAULT_BATCH_SIZE = 0;
```
npomaroli, line 20: We should have a default value, which enables batch processing.

api/src/main/java/com/gentics/mesh/etc/config/ContentConfig.java
```diff
@@ -14,19 +14,36 @@
 public class ContentConfig implements Option {
 
 	public static final String MESH_CONTENT_AUTO_PURGE_ENV = "MESH_CONTENT_AUTO_PURGE";
+	public static final String MESH_GRAPH_BATCH_SIZE_ENV = "MESH_GRAPH_BATCH_SIZE";
```
npomaroli, line 17: Name of the constant and name of the environment variable should be renamed to MESH_CONTENT_BATCH_SIZE, because this has nothing to do with "graph" (or does it?).

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/schema/impl/SchemaContainerVersionImpl.java
```diff
@@ -79,15 +83,32 @@ public Class<? extends HibSchema> getContainerClass() {
...
-	public Result<? extends NodeGraphFieldContainer> getDraftFieldContainers(String branchUuid) {
-		return new TraversalResult<>(toStream(mesh().database().getVertices(
-			NodeGraphFieldContainerImpl.class,
-			new String[] { SCHEMA_CONTAINER_VERSION_KEY_PROPERTY },
-			new Object[] { getUuid() })).filter(
-				v -> toStream(v.getEdges(Direction.IN, HAS_FIELD_CONTAINER))
-					.anyMatch(
-						e -> e.getProperty(BRANCH_UUID_KEY).equals(branchUuid) && ContainerType.get(e.getProperty(EDGE_TYPE_KEY)).equals(DRAFT)))
-				.map(v -> graph.frameElementExplicit(v, NodeGraphFieldContainerImpl.class)));
+	public Result<? extends NodeGraphFieldContainer> getDraftFieldContainers(String branchUuid, long offset, long limit) {
```
npomaroli, line 86: Parameter offset is not used, and should be removed (also from the API)

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/dao/impl/SchemaDaoWrapperImpl.java
```diff
@@ -140,8 +148,19 @@ public HibSchemaVersion findVersionByUuid(HibSchema schema, String versionUuid)
 	}
 
 	@Override
-	public Result<? extends HibNodeFieldContainer> findDraftFieldContainers(HibSchemaVersion version, String branchUuid) {
-		return toGraph(version).getDraftFieldContainers(branchUuid);
+	public long findDraftFieldContainerCount(HibSchemaVersion version, String branchUuid) {
```
npomaroli, line 151: This implementation will be inefficient (for large datasets) and it seems that this is only used to add some metrics. Maybe this method should be removed.

npomaroli (changes requested):

plyhun:
Moved to #1485

PR #1485, opened 2023-01-26
Chunked migration process

plyhun:
## Abstract

As of now the migration process tries to load all the data to process at once, which exhausts resources. We need to split the workload into chunks reliably.

### General

### On API Changes


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingContentDao.java
```diff
@@ -664,6 +664,10 @@ default void postfixPathSegment(HibNode node, String branchUuid, ContainerType t
 	private boolean updateWebrootPathInfo(HibNodeFieldContainer content, HibNode node, HibNodeFieldContainerEdge edge, String branchUuid, String segmentFieldName,
 										  String conflictI18n,
 										  ContainerType type) {
+		// Early return, if no segment field set
```
npomaroli, line 667: Should this be in here?

npomaroli (approved):

PR #1486, opened 2023-01-27
Cluster test fixes

plyhun:
## Abstract

Since recent OrientDB update the hard RAM limits have to be set for the cluster envs. The Java 8 environment is updated, since the old one is not anymore hosted.

### General

### On API Changes

### DIFF
--- a/test-common/src/main/java/com/gentics/mesh/test/docker/MeshContainer.java
+++ b/test-common/src/main/java/com/gentics/mesh/test/docker/MeshContainer.java
@@ -239,7 +239,7 @@ public class MeshContainer extends GenericContainer<MeshContainer> {
 			addEnv(ClusterOptions.MESH_CLUSTER_COORDINATOR_REGEX_ENV, coordinatorPlaneRegex);
 		}
 
-		String javaOpts = "-Dstorage.diskCache.bufferSize=1024 ";
+		String javaOpts = "-Dmemory.directMemory.preallocate=false ";
 		if (debugPort != null) {
 			javaOpts = "-agentlib:jdwp=transport=dt_socket,server=y,address=8000,suspend=n ";
 			exposedPorts.add(8000);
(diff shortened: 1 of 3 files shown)


test-common/src/main/resources/Dockerfile.local
```diff
@@ -1,4 +1,4 @@
-FROM    java:openjdk-8-jre-alpine
```
plyhun, line 1: I remember having recently some problems with using JRE image, so placed JDK here.

test-common/src/main/java/com/gentics/mesh/test/docker/MeshContainer.java
```diff
@@ -690,7 +691,7 @@ public String getContainerIpAddress() {
 		if (containerHost != null) {
 			return containerHost;
 		} else {
-			return super.getContainerIpAddress();
+			return DockerClientFactory.instance().dockerHostIpAddress();
```
plyhun, line 694: This makes a stack overflow.

test-common/src/main/java/com/gentics/mesh/test/docker/MeshContainer.java
```diff
@@ -238,7 +239,7 @@ protected void configure() {
 			addEnv(ClusterOptions.MESH_CLUSTER_COORDINATOR_REGEX_ENV, coordinatorPlaneRegex);
 		}
 
-		String javaOpts = null;
+		String javaOpts = "-Dmemory.directMemory.preallocate=false ";
```
npomaroli, line 242: This will get overwritten, if a debugPort has been set

test-common/src/main/java/com/gentics/mesh/test/docker/MeshContainer.java
npomaroli: Maybe it would be better to move the whole implementation to getHost() and do not overwrite the deprecated getContainerIpAddress() at all?

test-common/src/main/resources/Dockerfile.local
npomaroli: In the Dockerfile for the server, we use adoptopenjdk/openjdk11:x86_64-alpine-jdk-11.0.16.1_1 as base image, maybe it would be a good idea to use the same for testing

npomaroli (changes requested):

PR #1491, opened 2023-02-23
Add rsync to the server image

plyhun:
## Abstract

The rsync command is currently not available in the mesh container. For managing in openshift and backup purposes this is needed to reliably copy/backup the data from a container

### General

### On API Changes

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,5 +1,8 @@
 FROM adoptopenjdk/openjdk11:x86_64-alpine-jdk-11.0.16.1_1
 
+# Install dependencies
+RUN apk update && apk add rsync
+
 ENV MESH_AUTH_KEYSTORE_PATH=/keystore/keystore.jks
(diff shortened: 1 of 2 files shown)


server/Dockerfile
```diff
@@ -24,7 +24,8 @@ RUN adduser -D -u 1000 -G root -h /mesh mesh && \
     mkdir /mesh/data && \
     mkdir -p /mesh/elasticsearch/data && \
     chown 1000:0 /mesh -R && \
-    chmod 770 /mesh -R
+    chmod 770 /mesh -R && \
+    apk update && apk add rsync
```
bernhardriegler, line 28: I suggest to remove the apk cache after the adding the rsync program to keep the image size small
```suggestion
    apk update && apk add rsync && rm -rf /var/cache/apk/*
```

plyhun:
Close in favor of https://github.com/gentics/mesh/pull/1492 .

PR #1506, opened 2023-04-27
GPU-570 async virus check

yrucrem:
## Abstract

Adds support for external checks for uploaded binaries. When a check service URL is specified in a binary field of a schema, newly uploaded files will be marked as POSTPONED and a request to the configured URL is sent to initiate the check. The service is then expected to post the result to a callback endpoint to set the status to either ACCEPTED or DENIED. Only binaries with a status of ACCEPTED can be downloaded.

### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryDownloadHandler.java
```diff
@@ -78,12 +82,28 @@ public void handleReadBinaryField(RoutingContext rc, String uuid, String fieldNa
...
 					throw error(NOT_FOUND, "error_s3binaryfield_not_found_with_name", fieldName);
 				}
+
+				// The binary can only be downloaded if it has been accepted, or the current request provided the check
+				// secret (which should only be known to the check service).
+				if (field.getBinary().getCheckStatus() != BinaryCheckStatus.ACCEPTED
+						&& !Objects.equals(field.getBinary().getCheckSecret(), rc.queryParams().get("secret"))) {
+					throw error(NOT_FOUND, "error_binaryfield_not_accepted", fieldName);
+				}
+
```
plyhun, line 106: Can this copied part be externalized to an own function / closure?

plyhun (approved):
Approved + changes that we discussed yesterday.

PR #1516, opened 2023-07-03
SUP-15421: Move node publish to the separate nonlocked tx

plyhun:
## Abstract

In case of distributed Hazelcast cache, a deadlock may happen on large number of simultaneous file upload requests, caused by an attempt of deletion of an old content version, that has been locked by an upload process.

### General

### On API Changes


plyhun:
Obsolete.

PR #1517, opened 2023-07-06
Hotfix 1.10.x sup 15427

npomaroli:
### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/changelog/highlevel/HighLevelChangesList.java
+++ b/core/src/main/java/com/gentics/mesh/changelog/highlevel/HighLevelChangesList.java
@@ -16,14 +16,14 @@ import com.gentics.mesh.core.data.changelog.HighLevelChange;
 @Singleton
 public class HighLevelChangesList {
 
-	@Inject
-	public ExtractPlainText plainText;
+	protected final ExtractPlainText plainText;
+
+	protected final SetAdminUserFlag setAdminUserFlag;
 
 	@Inject
-	public SetAdminUserFlag setAdminUserFlag;
-
-	@Inject
-	public HighLevelChangesList() {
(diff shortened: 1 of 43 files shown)


core/src/main/java/com/gentics/mesh/changelog/highlevel/HighLevelChangesList.java
```diff
@@ -16,14 +16,14 @@
 @Singleton
 public class HighLevelChangesList {
 
-	@Inject
-	public ExtractPlainText plainText;
+	protected final ExtractPlainText plainText;
```
plyhun, line 19: Hm, shouldn't it be vice versa - the field injections **don't** care about order, but the constructor params injections **do** care?

npomaroli: The problem with the field injections is that dagger will create Factory classes that have methods for actually injecting the fields and the order of the parameters of this method is not necessarily the order of the fields in the class.

By using a constructor, the order of the parameters is well defined.

core/src/main/java/com/gentics/mesh/core/migration/AbstractMigrationHandler.java
```diff
@@ -160,7 +160,7 @@ protected <T> List<Exception> migrateLoop(Queue<T> containers, EventCauseInfo ca
 				errorsDetected.add(new MigrationAbortedException("Local node is not online."));
 				return errorsDetected;
 			}
-			if (clusteringEnabled && !delegator.isMaster()) {
```
plyhun, line 163: TL/DR: this is an actual fix.

plyhun (approved):

PR #1519, opened 2023-07-14
Add API for preloading list field values

npomaroli:
Add API for preloading list field values
Omit checking for conflicting url field values, if the values do not change.
Fix graphql test for unordered items (like tags) to not rely on the order.

### General

### On API Changes

### DIFF
--- a/tests/common/src/main/java/com/gentics/mesh/assertj/impl/JsonObjectAssert.java
+++ b/tests/common/src/main/java/com/gentics/mesh/assertj/impl/JsonObjectAssert.java
@@ -1,6 +1,7 @@
 package com.gentics.mesh.assertj.impl;
 
 import static com.gentics.mesh.MeshVersion.CURRENT_API_BASE_PATH;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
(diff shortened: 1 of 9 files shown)


tests/common/src/main/java/com/gentics/mesh/assertj/impl/JsonObjectAssert.java
```diff
@@ -80,8 +85,27 @@ public JsonObjectAssert hasNot(String path, String msg) {
 	public JsonObjectAssert has(String path, String value, String msg) {
 		try {
 			Object actualValue = getByPath(path);
-			String actualStringRep = String.valueOf(actualValue);
-			assertEquals("Value for property on path {" + path + "} did not match: " + msg, value, actualStringRep);
+			if (actualValue instanceof Collection<?>) {
```
npomaroli, line 88: This enhancement allows assertions for value sets (order of the values is irrelevant)

tests/tests-core/src/main/resources/graphql/node-tag-query
```diff
@@ -2,9 +2,7 @@
 	node(path:"/Products/Concorde.en.html") {
 		tags {
 			elements {
-				# [$.data.node.tags.elements[0].name=Plane]
-				# [$.data.node.tags.elements[1].name=Twinjet]
-				# [$.data.node.tags.elements[2].name=red]
+				# [$.data.node.tags.elements[*].name=[Plane, Twinjet, red]]
```
npomaroli, line 5: Tags are not sorted, so we check for the set (order is irrelevant)

plyhun (approved):

PR #1520, opened 2023-07-25
HTTP protocol version options for Mesh REST client

plyhun:
## Abstract

Allow HTTP protocol version of Mesh REST client being configurable.

### General

### On API Changes


rest-client/src/main/java/com/gentics/mesh/rest/client/impl/OkHttpClientUtil.java
```diff
@@ -125,6 +128,25 @@ private static void initializeHttpClient(Builder builder, MeshRestClientConfig c
 			builder.sslSocketFactory(sslCtx.getSocketFactory(), (X509TrustManager) trustManagers[0]);
 		}
 
+		if (config.getProtocolVersion() != null) {
+			switch (config.getProtocolVersion()) {
+			case DEFAULT:
```
plyhun, line 133: https://github.com/square/okhttp/blob/okhttp_3.12.x/okhttp/src/main/java/okhttp3/OkHttpClient.java#L127

npomaroli (approved):

PR #1521, opened 2023-07-25
ALPN

plyhun:
## Abstract

Config for ALPN, speeding up HTTP/2 connections.

### General

### On API Changes


core/src/main/java/com/gentics/mesh/rest/RestAPIVerticle.java
```diff
@@ -261,7 +261,8 @@ private void applyCommonSettings(HttpServerOptions options, HttpServerConfig ser
 		options.setHost(host);
 		options.setCompressionSupported(true);
 		options.setHandle100ContinueAutomatically(true);
-		// options.setLogActivity(true);
+		options.setUseAlpn(serverConfig.isUseAlpn());
+		options.setLogActivity(true);
```
npomaroli, line 265: Was this activated on purpose? What kind of log will this produce?

plyhun: No, it is a leak. Removing.

npomaroli (approved):

PR #1523, opened 2023-07-27
SUP-15549: Refactor caches. Add missing caches for named entities.

plyhun:
## Abstract

Cache all the named entities (except Nodes).

### General

### On API Changes


mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/dagger/module/OrientDBModule.java
```diff
@@ -277,6 +287,21 @@ public abstract class OrientDBModule {
 
 	// Caches
 
+	@Binds
+	abstract UserNameCache bindUserNameCache(UserNameCacheImpl e);
```
npomaroli, line 291: It seems that the new caches are never used.

npomaroli (changes requested):

npomaroli (approved):

PR #1532, opened 2023-08-30
Fix incorrect implementation of SortingParameters

npomaroli:
Add tests for sorting entity lists

### General

### On API Changes

### DIFF
--- a/rest-client/src/main/java/com/gentics/mesh/parameter/client/SortingParametersImpl.java
+++ b/rest-client/src/main/java/com/gentics/mesh/parameter/client/SortingParametersImpl.java
@@ -1,28 +1,6 @@
 package com.gentics.mesh.parameter.client;
 
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import com.gentics.mesh.core.rest.SortOrder;
 import com.gentics.mesh.parameter.SortingParameters;
 
 public class SortingParametersImpl extends AbstractParameters implements SortingParameters {
-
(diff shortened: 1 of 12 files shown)


rest-client/src/main/java/com/gentics/mesh/parameter/client/SortingParametersImpl.java
```diff
@@ -1,28 +1,6 @@
...
-import java.util.function.Function;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
-import com.gentics.mesh.core.rest.SortOrder;
 import com.gentics.mesh.parameter.SortingParameters;
 
 public class SortingParametersImpl extends AbstractParameters implements SortingParameters {
-
-	private final Map<String, SortOrder> sort = new LinkedHashMap<>();
```
plyhun, line 14: As the sorting can be used over several fields, including the ones not supported in the query line, this content has to be back.

plyhun: It is used [below](https://github.com/gentics/mesh/pull/1532/files#diff-49eddd22b8e20004a0cbe175434b5123f168fa26fe86286b4a866d6600f409c7R68) as well.

plyhun (approved):

rest-client/src/main/java/com/gentics/mesh/parameter/client/SortingParametersImpl.java
npomaroli: As mentioned in our talk: this class is only used with the MeshRestClient for passing parameters to requests and the removed code prevented this.

PR #1541, opened 2023-10-04
Per-schema/field configuration for Elasticsearch indexing

plyhun:
## Abstract

At the moment Mesh gives all the content to the indexing engine, which is a security issue, since not all the indexed data is insensible enough. This feature enables a `No Index` flag, attached to the selected (micro)schemas and fields, banning the corresponding content from being indexed.

### General

### On API Changes

### DIFF
--- a/elasticsearch/src/main/java/com/gentics/mesh/search/index/group/GroupIndexHandlerImpl.java
+++ b/elasticsearch/src/main/java/com/gentics/mesh/search/index/group/GroupIndexHandlerImpl.java
@@ -72,10 +72,10 @@ public class GroupIndexHandlerImpl extends AbstractIndexHandler<HibGroup> implem
 	}
 
 	@Override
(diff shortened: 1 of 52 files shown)


elasticsearch/src/main/java/com/gentics/mesh/search/index/group/GroupIndexHandlerImpl.java
```diff
@@ -72,10 +72,10 @@ public GroupMappingProvider getMappingProvider() {
 	}
 
 	@Override
-	public Map<String, IndexInfo> getIndices() {
+	public Map<String, Optional<IndexInfo>> getIndices() {
```
plyhun, line 75: The main point is to replace index creation with `IndexInfo` with an `Optional<IndexInfo>`, meaning the presence as creation, and the absence as removal/ignorance.

PR #1542, opened 2023-10-09
SUP-15904: Speed up the parents fetch

plyhun:
## Abstract

One can fetch the parent node along with the node and its content, to speed up GraphQL requests

### General

### On API Changes


plyhun:
Better solution comes along.

PR #1546, opened 2023-10-16
Remove annoying shard output

plyhun:
## Abstract

The System.out input heavily pollutes the log.

### General

### On API Changes


npomaroli (approved):

server/src/main/java/com/gentics/mesh/server/DevRunner.java
```diff
@@ -45,8 +45,6 @@ public static void main(String[] args) throws Exception {
 		defaultOption.getDebugInfoOptions().setLogEnabled(false);
 		OrientDBMeshOptions options = OptionsLoader.createOrloadOptions(OrientDBMeshOptions.class, defaultOption, args);
```
plyhun, line 47: This is utterly awkward. As a dev, I want the DevRunner to have the ability to start an indexer.

PR #1550, opened 2023-10-24
SUP-15948: Null aware Cache

plyhun:
## Abstract

Performance drawbacks come out of the current cache implementation, that treats the null value as cache miss.

### General

### On API Changes


common/src/main/java/com/gentics/mesh/cache/impl/EventAwareCacheImpl.java
```diff
@@ -186,9 +190,9 @@ public V get(K key, Function<? super K, ? extends V> mappingFunction) {
 			} else {
 				missCounter.increment();
 			}
-			return value;
+			return value.orElse(null);
 		} else {
-			return cache.get(key, mappingFunction);
+			return cache.get(key, (k) -> Optional.ofNullable(mappingFunction.apply(k))).orElse(null);
```
plyhun, line 195: This should be aligned with the latest cache updates.

common/src/main/java/com/gentics/mesh/cache/impl/EventAwareCacheImpl.java
```diff
@@ -35,7 +38,7 @@ public class EventAwareCacheImpl<K, V> implements EventAwareCache<K, V> {
 
 	private static final Logger log = LoggerFactory.getLogger(EventAwareCacheImpl.class);
 
-	private final Cache<K, V> cache;
+	private final Cache<K, Optional<V>> cache;
```
npomaroli, line 41: Most implementation of AbstractMeshCache (which internally uses EventAwareCache) will not clean the cache upon creation of elements. This means that cached empty Optionals will not be invalidated properly.
We definitely need test cases for ALL implementations for (example made for project name):
1. Load (inexistent) project by name (will now put an empty Optional into the cache)
2. Create project by that name
3. Try to load the project with the name again (should find the project now)

npomaroli (changes requested):

common/src/main/java/com/gentics/mesh/cache/impl/EventAwareCacheImpl.java
plyhun: Added the reaction over `*CREATED` events. The unit test set is TBD.

mdm/api/src/main/java/com/gentics/mesh/core/data/Taggable.java
```diff
@@ -93,7 +93,6 @@ default List<HibTag> getTagsToSet(List<TagReference> list, InternalActionContext
 					if (userDao.hasPermission(user, tagFamily, CREATE_PERM)) {
 						tag = tagDao.create(tagFamily, tagReference.getName(), project, user);
 						userDao.inheritRolePermissions(user, tagFamily, tag);
-						batch.add(tag.onCreated());
```
plyhun, line 96: Here and below - `onCreated()` moved into `recache()` method below.

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -12,13 +13,20 @@
...
 		HibLanguage language = createPersisted(uuid);
 		language.setName(languageName);
 		language.setLanguageTag(languageTag);
-		return mergeIntoPersisted(language);
+		mergeIntoPersisted(language);
+
+		maybeGetCache().ifPresent(cache -> {
+			cache.clear(languageName);
+			cache.clear(languageTag);
+			CommonTx.get().data().mesh().batchProvider().get().add(language.onCreated());
```
plyhun, line 27: Here and below - we clear the cache explicitly, and sending the creation event to notify the cluster to clear it as well.

common/src/main/java/com/gentics/mesh/cache/impl/EventAwareCacheImpl.java
plyhun: UTs added.

tests/tests-core/src/main/java/com/gentics/mesh/core/NamedEntityDuplicationTest.java
```diff
@@ -21,44 +21,49 @@ public class NamedEntityDuplicationTest extends AbstractMeshTest {
...
+		testDuplicate("group", (name, tx) -> tx.groupDao().create(name, user()));
 	}
 
 	@Test(expected = RuntimeException.class)
 	public void testDuplicateLanguage() {
-		testDuplicate("language", (name, tx) -> Tx.get().languageDao().create(name, "lng"));
+		testDuplicate("language", (name, tx) -> tx.languageDao().create(name, "lng"));
+	}
+
+	@Test(expected = RuntimeException.class)
```
plyhun, line 57: Forgotten test (passes).

tests/tests-core/src/main/java/com/gentics/mesh/core/NamedEntityDuplicationTest.java
```diff
@@ -21,44 +21,49 @@ public class NamedEntityDuplicationTest extends AbstractMeshTest {
 
 	@Test(expected = RuntimeException.class)
 	public void testDuplicateProject() {
-		testDuplicate("project", (name, tx) -> Tx.get().projectDao().create(name, "localhost", false, "", user(), schemaContainer("folder").getLatestVersion(), data().createBatch()));
+		testDuplicate("project", (name, tx) -> tx.projectDao().create(name, "localhost", false, "", user(), schemaContainer("folder").getLatestVersion(), data().createBatch()));
```
plyhun, line 24: Small refactor.

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingNamedEntityDao.java
```diff
@@ -0,0 +1,46 @@
...
+
+	/**
+	 * Chear the cached name, along with sending the specified event.
+	 * 
+	 * @param entity
+	 * @param event
+	 */
+	default void recache(T entity, MeshEventModel event) {
+		uncache(entity);
+		Tx.maybeGet().map(CommonTx.class::cast).map(ctx -> ctx.data().mesh().batchProvider()).ifPresent(bp -> bp.get().add(event));
```
npomaroli, line 44: This will not work as intended. This will just get a new instance of EventQueueBatch and will add the event to it, but since dispatch() is never called for this instance, the events will just be lost.

npomaroli (changes requested):

core/src/main/java/com/gentics/mesh/core/endpoint/auth/AuthenticationRestHandler.java
```diff
@@ -42,7 +43,9 @@ public AuthenticationRestHandler(MeshJWTAuthProvider authProvider, HandlerUtilit
 	public void handleMe(InternalActionContext ac) {
 		utils.syncTx(ac, tx -> {
 			// TODO add permission check
-			HibUser requestUser = CommonTx.get().userDao().mergeIntoPersisted(ac.getUser());
+			PersistingUserDao userDao = CommonTx.get().userDao();
+			HibUser requestUser = userDao.mergeIntoPersisted(ac.getUser());
```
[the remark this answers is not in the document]
plyhun: Side effect of `mergeIntoPersisted` is re-attaching a new/stale entity into the ORM, so all its fields are accessible. Perhaps we heed an API wide `refresh()` method.

npomaroli (changes requested):

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingMicroschemaDao.java
```diff
@@ -229,7 +229,9 @@ default HibMicroschema create(MicroschemaVersionModel microschema, HibUser user,
 		container.setName(microschema.getName());
 		container.generateBucketId();
 
-		return mergeIntoPersisted(container);
+		uncacheAsync(mergeIntoPersisted(container), container.onCreated());
+		uncacheSync(container);
+		return container;
```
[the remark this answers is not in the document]
plyhun: The entity is actually the same in inparam and out. Though I'll move it around for clarity.

tests/tests-core/src/main/java/com/gentics/mesh/test/context/MeshTestContext.java
```diff
@@ -171,6 +169,8 @@ public void setup(MeshTestSetting settings) throws Exception {
 		default:
 			break;
 		}
+		boolean setAdminPassword = settings.optionChanger() != MeshCoreOptionChanger.INITIAL_ADMIN_PASSWORD;
```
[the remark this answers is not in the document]
plyhun: After each distinct test we clean up. The cleanup includes the indices wipe, but does not restart Mesh. So on the second test we have the running Mesh, and no data. When we start setting the data up, Mesh sees the ES as ready and sends all the created entities to it (this is happening because of the events processing happening on each succeeded Tx, and not manually as before). With no index definition yet existing in ES, ES starts generating the definitions, based on the data itself, which differs from the ones being explicitly sent by the setupIndexHandlers() request, and as a consequence - every second ES-related test fails because of wrong request body JSON.
Yes, this was a tough one to detect.

npomaroli (approved):

elasticsearch/src/main/java/com/gentics/mesh/search/index/node/NodeIndexHandlerImpl.java
```diff
@@ -633,4 +633,9 @@ public Stream<? extends HibNode> loadAllElements() {
 		return CommonTx.get().nodeDao().findAllGlobal();
 	}
 
+	@Override
+	public boolean isDefinitionDataDependent() {
```
plyhun, line 637: Here and below: there is a need to split the test context init into the data-unaware index init (comes before the data population) , and data-aware index init (after the data is populated).

tests/common/src/main/java/com/gentics/mesh/assertj/impl/DummySearchProviderAssert.java
```diff
@@ -38,24 +41,24 @@ public DummySearchProviderAssert(TrackingSearchProvider actual) {
 	}
 
 	public DummySearchProviderAssert recordedStoreEvents(int count) {
-		isNotNull();
-		String info = actual.getStoreEvents().keySet().stream().map(Object::toString).reduce((t, u) -> t + "\n" + u).orElse("");
-		assertEquals("The search provider did not record the correct amount of store events. Found events: {\n" + info + "\n}", count, actual
-				.getStoreEvents().size());
-		return this;
+		return recordedEvents(p -> p.getStoreEvents().keySet(), takeAll -> true, count);
```
plyhun, line 44: The event assertion API is a bit generalized.

tests/tests-core/src/main/java/com/gentics/mesh/test/context/AbstractMeshTest.java
```diff
@@ -78,6 +78,7 @@ public MeshTestContext getTestContext() {
 	@Before
 	public void setupEventAsserter() {
 		eventAsserter = new EventAsserter(getTestContext());
+		testContext.waitAndClearSearchIdleEvents();
```
plyhun, line 81: The events, triggered by the initial data, have to be cleared, to prevent polluting the target event tracking.

tests/tests-core/src/main/java/com/gentics/mesh/test/context/MeshTestContext.java
```diff
@@ -160,13 +160,25 @@ protected void starting(Description description) throws Throwable {
 	public void setup(MeshTestSetting settings) throws Exception {
 		meshTestContextProvider.getInstanceProvider().initMeshData(settings, meshDagger);
 		initFolders(mesh.getOptions());
+		listenToSearchIdleEvent();
+		// Set up the index definition of data independent entities.
+		switch (settings.elasticsearch()) {
+		case CONTAINER_ES6:
+		case CONTAINER_ES7:
```
plyhun, line 167: Data-unaware vs -aware index init.

PR #1555, opened 2023-11-07
GPU-1088: Get session token from API token

plyhun:
### General

### On API Changes

### DIFF
--- a/common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
+++ b/common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
@@ -8,9 +8,6 @@ import static io.netty.handler.codec.http.HttpResponseStatus.UNAUTHORIZED;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
-import com.gentics.mesh.auth.handler.MeshJWTAuthHandler;
-import io.vertx.core.http.Cookie;
-import io.vertx.ext.auth.authentication.AuthenticationProvider;
 import org.apache.commons.lang.NotImplementedException;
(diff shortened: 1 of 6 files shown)


common/src/main/java/com/gentics/mesh/auth/provider/MeshJWTAuthProvider.java
```diff
@@ -305,4 +303,33 @@ public void login(InternalActionContext ac, String username, String password, St
...
+	/**
+	 * Handle the login action and set a token cookie if the apiToken is valid.
+	 *
+	 * @param ac
+	 *            Action context used to add token cookie
+	 * @param apiToken
+	 *            Mesh API token
+	 */
+	public void login(InternalActionContext ac, String apiToken) {
+		JsonObject authInfo = new JsonObject().put("token", apiToken).put("options", new JsonObject());
```
plyhun, line 315: The usage is analogous to the one with `apiToken` pulled out of auth header.

tests/tests-core/src/main/java/com/gentics/mesh/core/user/AuthenticationEndpointTest.java
```diff
@@ -67,6 +70,32 @@ public void testRestClient() throws Exception {
 		}
 	}
 
+	@Test
+	public void testApiTokenLogin() {
+		UserResponse response = call(() -> client().me());
+		String token = call(() -> client().issueAPIToken(response.getUuid())).getToken();
+		client().setAPIKey(token).setLogin(null, null, null);
```
plyhun, line 77: An example of Mesh REST client usage.

yrucrem (approved):
Looks good to me.

PR #1556, opened 2023-11-08
SUP-16041: Name uniqueness tests

plyhun:
## Abstract

A set of uniqueness checks against selected named entities has been introduced.

### General

### On API Changes

### DIFF
--- a/tests/tests-core/src/main/java/com/gentics/mesh/core/NamedEntityDuplicationTest.java
+++ b/tests/tests-core/src/main/java/com/gentics/mesh/core/NamedEntityDuplicationTest.java
@@ -1,0 +1,84 @@
+package com.gentics.mesh.core;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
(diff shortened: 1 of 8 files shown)


tests/tests-core/src/main/java/com/gentics/mesh/core/NamedEntityDuplicationTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+	protected <E extends HibNamedElement> void testDuplicate(String entityName, BiFunction<String, Tx, E> creator) {
+		String name = entityName + "_" + Long.toHexString(System.currentTimeMillis());
+
+		tx(tx -> {
+			E role = creator.apply(name, tx);
+			assertNotNull(role);
+			assertEquals(name, role.getName());
+
+			role = creator.apply(name, tx);
```
plyhun, line 81: Enterprise prior to fixes crashes almost everything. ODB crashes the tag-related checks only. 
Enterprise with the fixes passes.

PR #1557, opened 2023-11-10
SUP-16040: Batch load the referencedBy nodes

plyhun:
## Abstract

Speed up the node reference fields by batching the load.

### General

### On API Changes

### DIFF
--- a/verticles/graphql/src/main/java/com/gentics/mesh/graphql/model/NodeReferenceIn.java
+++ b/verticles/graphql/src/main/java/com/gentics/mesh/graphql/model/NodeReferenceIn.java
@@ -3,11 +3,19 @@ package com.gentics.mesh.graphql.model;
 import static com.gentics.mesh.core.rest.common.ContainerType.DRAFT;
 import static com.gentics.mesh.core.rest.common.ContainerType.PUBLISHED;
 
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import org.apache.commons.lang3.tuple.Pair;
+
 import com.gentics.graphqlfilter.util.Lazy;
-import com.gentics.mesh.core.data.HibNodeFieldContainer;
 import com.gentics.mesh.core.data.dao.ContentDao;
+import com.gentics.mesh.core.data.node.HibNode;
 import com.gentics.mesh.core.data.node.NodeContent;
 import com.gentics.mesh.core.data.node.field.nesting.HibNodeField;
 import com.gentics.mesh.core.db.Tx;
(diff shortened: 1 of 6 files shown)


verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/NodeTypeProvider.java
```diff
@@ -407,16 +408,29 @@ public Versioned<List<GraphQLFieldDefinition>> createNodeInterfaceFields(GraphQL
...
-					return new DynamicStreamPageImpl<>(stream, getPagingInfo(env));
-				}).build(),
+					DataLoader<NodeContent, Collection<NodeReferenceIn>> parentLoader = env.getDataLoader(NodeDataLoader.REFERENCED_BY_LOADER_KEY);
+					return parentLoader.load(content).thenApply(c -> {
+						Stream<NodeReferenceIn> stream = c.stream();
+						Map<String, ?> filterInput = env.getArgument("filter");
+						if (filterInput != null) {
+							stream = stream.filter(nodeReferenceFilter(context).createPredicate(filterInput));
+						}
+						return new DynamicStreamPageImpl<>(stream, getPagingInfo(env));
```
plyhun, line 430: This surprisingly works, the results match with the old version.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/NodeTypeProvider.java
```diff
@@ -407,16 +407,29 @@ public Versioned<List<GraphQLFieldDefinition>> createNodeInterfaceFields(GraphQL
...
-					if (filterInput != null) {
-						stream = stream.filter(nodeReferenceFilter(context).createPredicate(filterInput));
+					if (content == null) {
+						return null;
+					}
+					HibNode node = content.getNode();
+					if (node == null) {
+						return null;
 					}
+					List<String> languageTags = getLanguageArgument(env, content);
```
npomaroli, line 419: This is not needed here.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/NodeTypeProvider.java
```diff
@@ -407,16 +407,29 @@ public Versioned<List<GraphQLFieldDefinition>> createNodeInterfaceFields(GraphQL
...
+					ContainerType type = getNodeVersion(env);
```
npomaroli, line 420: This is not used anymore, which changes the behaviour.
In the original implementation, the type was take from the environment, the batch loader will now use the type of each content to load the referencing node contents.

npomaroli: This can be tested with
`referencedBy(version: draft)` vs. `referencedBy(version: published)`

verticles/graphql/src/main/java/com/gentics/mesh/graphql/model/NodeReferenceIn.java
```diff
@@ -58,6 +66,43 @@ public static Stream<NodeReferenceIn> fromContent(GraphQLContext gc, NodeContent
...
+		String branchUuid = tx.getBranch(gc).getUuid();
+
+		// content per node
+		Map<HibNode, List<NodeContent>> nodeContent = contents.stream().collect(Collectors.groupingBy(NodeContent::getNode, Collectors.mapping(Function.identity(), Collectors.toList())));
+
+		// 1. Collecting referencing fields for the nodes
+		return contentDao.getInboundReferences(
+					contents.stream().map(NodeContent::getNode).collect(Collectors.toSet())
+				// 2. Collect the contents of the referencing fields
+				).flatMap(refNode -> refNode.getKey().getReferencingContents()
```
npomaroli, line 87: This will still use extra queries per node.

npomaroli (changes requested):

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/NodeTypeProvider.java
plyhun: Fixed.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/model/NodeReferenceIn.java
npomaroli: I tested the current implementation by loading "referencedBy" for 10 nodes, where each one is referenced by 10 other nodes. The REFERENCED_BY_LOADER used over 600 SELECT Statements for that.

plyhun: Actual for Enterprise only -> goes into Enterprise part.

npomaroli (approved):

PR #1559, opened 2023-11-21
SUP-16000: Fail-safe the GraphQL calls with the (micro)schemas with no fields

plyhun:
## Abstract

Some GraphQL requests rely on every (micro)schema having at least one fields, which may not be the case.

### General

### On API Changes


verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/MicroschemaTypeProvider.java
```diff
@@ -116,13 +108,7 @@ public GraphQLObjectType createType() {
...
 
 		// .fields
-		schemaType.field(newFieldDefinition().name("fields").type(type).dataFetcher(env -> {
-			List<FieldSchema> fields = loadModelWithFallback(env).getFields();
-			if (fields.isEmpty()) {
-				fields = Collections.singletonList(emptySchemaFieldDummy("This microschema has no fields. Do not use this filter."));
-			}
-			return fields;
-		}));
+		schemaType.field(newFieldDefinition().name("fields").type(type).dataFetcher(env -> loadModelWithFallback(env).getFields()));
```
plyhun, line 111: This is safe to remove, since we return a list of not the GraphQLType, but FieldSchema, which is legal.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/NodeTypeProvider.java
```diff
@@ -198,7 +199,10 @@ public Versioned<GraphQLType> createType(GraphQLContext context) {
 			nodeType.description(
 				"A Node is the basic building block for contents. Nodes can contain multiple language specific contents. These contents contain the fields with the actual content.");
 			interfaceTypeProvider.addCommonFields(nodeType, true);
-			nodeType.fields(createNodeInterfaceFields(context).forVersion(context));
+			List<GraphQLFieldDefinition> fields = createNodeInterfaceFields(context).forVersion(context);
```
plyhun, line 202: Here and below - create no `fields` type, if the target (micro)schema has no fields.

npomaroli (approved):

PR #1562, opened 2023-11-27
SUP-15674: Disallow empty or spaced password

plyhun:
## Abstract

It is possible ATM to set an empty string as password, that should not be the case.

### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/core/user/UserEndpointTest.java
```diff
@@ -1005,6 +1005,29 @@ public void testCreateUserWithMissingUuidNameInNodeReference() {
 	//
 	// }
 
+	@Test
+	public void testUpdatePasswordWithEmptyPassword() {
```
npomaroli, line 1009: I don't understand this test.
First, the password hash is set to null, but the transaction is rolled back (because the call to tx.success() is missing), so the stored password hash is still the original one, not null, which is stored in oldHash.
Then a request to set the password to a blank string is done (which should not change anything)
And the assertion asserts, that the password hash of the user (which was still not changed) is not equal to null.

npomaroli (changes requested):

npomaroli (approved):

PR #1563, opened 2023-11-29
SUP-16120: Fix the instances initialization

plyhun:
## Abstract

At 1.10.20 the regression appeared, introducing the problems at creating a new named entity. Th e fix comes here.

### General

### On API Changes

### DIFF
--- a/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingDao.java
+++ b/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingDao.java
@@ -1,5 +1,7 @@
 package com.gentics.mesh.core.data.dao;
 
+import java.util.function.Consumer;
+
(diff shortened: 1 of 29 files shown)


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingDao.java
```diff
@@ -24,9 +26,11 @@ public interface PersistingDao<T extends HibBaseElement> {
 	 * 
 	 * @param uuid
 	 *            Optional uuid
+	 * @param inflater
+	 *            The instance inflater
 	 * @return
 	 */
-	T createPersisted(String uuid);
+	T createPersisted(String uuid, Consumer<T> inflater);
```
plyhun, line 33: Main change.

npomaroli (approved):

PR #1564, opened 2023-12-01
Languages Endpoint

plyhun:
## Abstract

An endpoint for languages retrieval.

### General

### On API Changes


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingProjectDao.java
```diff
@@ -122,6 +128,9 @@ default HibProject create(String name, String hostname, Boolean ssl, String path
 		HibProject project = createPersisted(uuid);
 		project.setName(name);
 
+		// add the default language
```
plyhun, line 131: Achtung: do we need to control this?

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/changelog/highlevel/change/AssignLanguagesToProject.java
```diff
@@ -0,0 +1,80 @@
...
+import com.gentics.mesh.core.db.CommonTx;
+import com.gentics.mesh.etc.config.MeshOptions;
+import com.gentics.mesh.graphdb.spi.GraphDatabase;
+import com.gentics.mesh.util.StreamUtil;
+
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
+@Singleton
+public class AssignLanguagesToProject extends AbstractHighLevelChange {
```
npomaroli, line 31: This should not be OrientDB specific.

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/changelog/highlevel/change/AssignLanguagesToProject.java
```diff
@@ -0,0 +1,80 @@
...
+	@Override
+	public String getDescription() {
+		return "Checks the languages existing in the project content, and assigns them to the project";
+	}
+
+	@Override
+	public void apply() {
+		CommonTx ctx = CommonTx.get();
+		PersistingLanguageDao languageDao = ctx.languageDao();
+		Map<HibProject, Set<String>> projectLangs = ctx.projectDao().findAll().stream()
```
npomaroli, line 66: This does not appear to be very efficient: it loads all contents for all nodes for all projects.
Especially when implementing this for Mesh SQL, we need to find a better way to do this.

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingNodeDao.java
```diff
@@ -2052,6 +2070,8 @@ default void addReferenceUpdates(HibNode updatedNode, BulkActionContext bac) {
 			});
 	}
 
+	Stream<? extends HibNodeFieldContainerEdge> findLanguageEdges(HibProject project, Collection<String> languageTags);
```
npomaroli, line 2073: This is only used to determine, whether any of the given languages is currently used in the project.
It is not necessary to return a Stream of all ContainerEdges here, maybe a map of language tag to boolean would be sufficient.
This way, we could find more efficient ways to determine the required information.

npomaroli (changes requested):

mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/changelog/highlevel/change/AssignLanguagesToProject.java
plyhun: SQL will hold an entirely different implementation.

plyhun:
-> https://github.com/gentics/mesh/pull/1578

PR #1569, opened 2023-12-27
Reorganize local storing of Tx

plyhun:
## Abstract

One should never store a thing in ThreadLocal, since we operate async contexts, not threads.

### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryUploadHandlerImpl.java
```diff
@@ -214,9 +216,12 @@ public void handleUpdateField(InternalActionContext ac, String nodeUuid, String
 				if (log.isDebugEnabled()) {
 					log.debug("Error detected. Purging previously stored upload for tempId {}", tmpId, e);
 				}
-				return binaryStorage.purgeTemporaryUpload(tmpId).doOnError(e1 -> {
-					log.error("Error while purging temporary upload for tempId {}", tmpId, e1);
-				}).onErrorComplete().andThen(Single.error(e));
+				return binaryStorage.purgeTemporaryUpload(tmpId)
+						// The file has already been deleted = not an error
```
plyhun, line 220: The reason for playing with tests is that after the fix the tests run **actually** asynchronous, and cause races in the handler.

mdm/api/src/main/java/com/gentics/mesh/core/db/Tx.java
```diff
@@ -23,22 +27,32 @@ public interface Tx extends BaseTransaction, DaoCollection, CacheCollection, Sec
...
+	 * Set the nested active transaction for the current Vert.x context, or thread.
 	 * 
 	 * @param tx
 	 *            Transaction
 	 */
 	static void setActive(Tx tx) {
-		Tx.threadLocalGraph.set(tx);
+		Optional.ofNullable(Vertx.currentContext())
+			.ifPresentOrElse(
+					ctx -> ctx.putLocal("tx", Optional.ofNullable(tx)),
```
plyhun, line 38: The actual change. There are `get()` and `getLocal()` in the Vert.x `Context`, which seem to be functionally equal, but `getLocal()` has internal usages.

mdm/api/src/main/java/com/gentics/mesh/core/db/Tx.java
```diff
@@ -23,22 +27,32 @@ public interface Tx extends BaseTransaction, DaoCollection, CacheCollection, Sec
...
-	 * Return the current active transaction. A transaction should be the only place where this threadlocal is updated.
+	 * Return the current active transaction. A transaction should be the only place where this context/threadlocal is updated.
 	 * 
 	 * @return Currently active transaction
 	 */
 	static Tx get() {
-		return Tx.threadLocalGraph.get();
+		Optional<Context> maybeVertxContext = Optional.ofNullable(Vertx.currentContext());
+		if (maybeVertxContext.isPresent()) {
+			Context vertxContext = maybeVertxContext.get();
```
plyhun, line 51: Don't mix the cases where we have an empty Context, and where we have no Context at all.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/binary/AbstractBinaryFieldUploadEndpointParameterizedTest.java
```diff
@@ -81,6 +81,8 @@ protected void testParallelUpload(boolean useSameName) throws IOException {
 		initialSyncWrites = initialSyncWrites.or(() -> Optional.of(mesh().globalLock().isSyncWrites()));
 		getTestContext().getInstanceProvider().setSyncWrites(isSyncWrites());
 
+		String defaultFilename = Long.toHexString(System.currentTimeMillis()) + "blume.jpg";
```
plyhun, line 84: Tell the file names apart per each test case.

tests/tests-core/src/main/java/com/gentics/mesh/core/graphql/GraphQLEndpointStressTest.java
```diff
@@ -0,0 +1,91 @@
...
+
+	protected final int simultaneousCalls;
+
+	protected MeshRestClient clientV2;
+	protected MeshRestClient clientV1;
+
+	@Parameters(name = "calls={0}")
+	public static Collection<Integer> paramData() {
+		int total = (int) streamQueries().count();
+		return List.of(1, 2, 3, 5, 8, 13, 21, total, 100, 250, 500, 1000);
```
plyhun, line 39: Run up to 1000 calls at once, to make Vert.x threads switch the native threads. Not always works :(

plyhun:
Postponed by now.

PR #1570, opened 2023-12-29
Hotfix 1.10.x sup 15643

plyhun:
## Abstract

Small fixes in `fields` REST parameter.

### General

### On API Changes

### DIFF
--- a/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
+++ b/mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
@@ -34,8 +34,7 @@ public interface PersistingLanguageDao extends LanguageDao, PersistingDaoGlobal<
 		FieldsSet fields = generic.getFields();
 
 		LanguageResponse model = new LanguageResponse();
-		// TODO align fields.isEmpty() with other DAOs on a major update - this is breaking at the moment.
-		if (fields.isEmpty() || fields.has("uuid")) {
+		if (fields.has("uuid")) {
 			model.setUuid(element.getUuid());
 
 			// Performance shortcut to return now and ignore the other checks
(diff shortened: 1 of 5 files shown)


mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -26,11 +30,38 @@ default HibLanguage create(String languageName, String languageTag, String uuid)
...
+		}
+		if (fields.isEmpty() || fields.has("name")) {
+			model.setName(element.getName());
+
+			// Performance shortcut to return now and ignore the other checks
+			if (fields.size() == 1) {
+				return model;
+			}
+		}
+		if (fields.isEmpty() || fields.has("tag")) {
```
npomaroli, line 54: Wouldn't it be better to align the possible values for "fields" with the attribute names in the rest model? We should use "languageTag" here instead of "tag".

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -26,11 +30,38 @@ default HibLanguage create(String languageName, String languageTag, String uuid)
...
+		GenericParameters generic = ac.getGenericParameters();
+		FieldsSet fields = generic.getFields();
+
 		LanguageResponse model = new LanguageResponse();
-		model.setUuid(element.getUuid());
-		model.setLanguageTag(element.getLanguageTag());
-		model.setName(element.getName());
-		model.setNativeName(element.getNativeName());
+		// TODO align fields.isEmpty() with other DAOs on a major update - this is breaking at the moment.
+		if (fields.isEmpty() || fields.has("uuid")) {
```
npomaroli, line 38: The check for fields.isEmpty() is not necessary, because fields.has(...) will return true if no fields restriction has been set anyways.

npomaroli (changes requested):

npomaroli (approved):

PR #1574, opened 2024-01-10
GraphQL Regression fixes

plyhun:
## Abstract

Since we are more careful with creating (micro)node types with no (micro)schemas, the references to these types should also be taken over.

### General

### On API Changes

### DIFF
--- a/verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/NodeFilter.java
+++ b/verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/NodeFilter.java
@@ -6,7 +6,6 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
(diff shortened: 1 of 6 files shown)


verticles/graphql/src/main/java/com/gentics/mesh/graphql/filter/NodeFilter.java
```diff
@@ -67,8 +66,8 @@ private NodeFilter(GraphQLContext context) {
 	private Optional<MainFilter<NodeContent>> createAllFieldFilters() {
 		HibProject project = Tx.get().getProject(context);
 		SchemaDao schemaDao = Tx.get().schemaDao();
-		List<FilterField<NodeContent, ?>> schemaFields = StreamSupport
-			.stream(schemaDao.findAll(project).spliterator(), false)
+		List<FilterField<NodeContent, ?>> schemaFields = schemaDao.findAll(project)
```
plyhun, line 69: Useless wrapper.

PR #1578, opened 2024-01-25
 Languages Endpoint

plyhun:
### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/actions/impl/LanguageDAOActionsImpl.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	@Override
+	public HibLanguage create(Tx tx, InternalActionContext ac, EventQueueBatch batch, String uuid) {
+		throw error(BAD_REQUEST, "error_language_creation_forbidden");
+	}
+
+	@Override
+	public HibLanguage loadByUuid(DAOActionContext ctx, String uuid, InternalPermission perm, boolean errorIfNotFound) {
+		return ctx.tx().languageDao().findByUuid(uuid);
```
npomaroli, line 38: This ignores errorIfNotFound, which causes subsequent NPEs when language endpoints are called with incorrect uuid's

core/src/main/java/com/gentics/mesh/core/actions/impl/LanguageDAOActionsImpl.java
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	@Override
+	public HibLanguage loadByName(DAOActionContext ctx, String name, InternalPermission perm, boolean errorIfNotFound) {
+		return ctx.tx().languageDao().findByName(name);
```
npomaroli, line 43: This ignores errorIfNotFound, which causes subsequent NPEs when language endpoints are called with incorrect language tags

tests/tests-core/src/main/java/com/gentics/mesh/core/language/LanguageEndpointTest.java
```diff
@@ -0,0 +1,368 @@
...
+import com.gentics.mesh.test.MeshTestSetting;
+import com.gentics.mesh.test.TestSize;
+import com.gentics.mesh.test.context.AbstractMeshTest;
+import com.gentics.mesh.test.definition.BasicRestTestcases;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+import io.reactivex.Observable;
+
+@MeshTestSetting(elasticsearch = TRACKING, testSize = TestSize.FULL, startServer = true)
+public class LanguageEndpointTest extends AbstractMeshTest implements BasicRestTestcases {
```
npomaroli, line 38: Even if creation/updating and deleting languages over the REST API is not supported, we should still add tests for the correct behaviour when the endpoints are called.

rest-model/src/main/java/com/gentics/mesh/core/rest/node/NodeCreateRequest.java
```diff
@@ -42,6 +42,10 @@ public class NodeCreateRequest implements FieldContainer {
 	@JsonPropertyDescription("Whether the publish the node after creation.")
 	private boolean publish = false;
 
+	@JsonProperty(required = false)
+	@JsonPropertyDescription("Assign a requested language to the project, if none was so far.")
+	private boolean assignLanguage = false;
```
npomaroli, line 47: The default value for this is not mentioned in the documentation and the default value is different from the default value of NodeUpdateRequest.
imho, both should be "true" (to reflect the "old" behaviour)

core/src/main/java/com/gentics/mesh/core/actions/impl/LanguageDAOActionsImpl.java
```diff
@@ -0,0 +1,93 @@
...
+@Singleton
+public class LanguageDAOActionsImpl implements LanguageDAOActions {
+
+	@Inject
+	public LanguageDAOActionsImpl() {
+	}
+
+	@Override
+	public HibLanguage create(Tx tx, InternalActionContext ac, EventQueueBatch batch, String uuid) {
+		throw error(BAD_REQUEST, "error_language_creation_forbidden");
```
npomaroli, line 33: Trying to create a language is not a "Bad Request", but should return a 405 (Method not allowed)

core/src/main/java/com/gentics/mesh/core/endpoint/project/LanguageEndpoint.java
```diff
@@ -39,22 +43,61 @@ public String getDescription() {
...
-			// Unassign languages should cause a batch process that removes the FieldContainers for the given language.
-			throw new NotImplementedException("not implemented");
-		});
-
-		Route getRoute = route("/:projectUuid/languages").method(GET).produces(APPLICATION_JSON);
-		getRoute.handler(rc -> {
-			throw new NotImplementedException("not implemented");
-		});
-	}
+		InternalEndpointRoute createRoute = createRoute();
```
npomaroli, line 46: Does it make sense to add the endpoints for creating/deleting, which will always return errors?
If the answer is yes, shouldn't we add endpoints for updating as well (also returning errors?)
imho, we should remove the unsupported endpoints completely.

npomaroli (changes requested):

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -23,6 +35,42 @@ default HibLanguage create(String languageName, String languageTag, String uuid)
...
+	}
+
+	@Override
+	default void unassign(HibLanguage language, HibProject project, EventQueueBatch batch, boolean throwOnInexisting) {
+		project.getLanguages().stream()
+			.filter(l -> l.getLanguageTag().equals(language.getLanguageTag()))
+			.findAny()
+			.ifPresentOrElse(existing -> {
+					CommonTx.get().nodeDao().findUsedLanguages(project, Collections.singletonList(language.getLanguageTag()), true).stream().findAny().ifPresentOrElse(existingContent -> {
+						throw error(BAD_REQUEST, "error_language_still_in_use", language.getLanguageTag(), project.getName());
```
npomaroli, line 61: Isn't CONFLICT the better response code here?

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -23,6 +35,42 @@ default HibLanguage create(String languageName, String languageTag, String uuid)
...
+					}, () -> {
+						project.removeLanguage(language);
+					});
+				}, () -> {
+					if (throwOnInexisting) {
+						throw error(BAD_REQUEST, "error_language_not_assigned", language.getLanguageTag(), project.getName());
```
npomaroli, line 67: Isn't NOT_FOUND the better response code here?
Or should we remove this, because throwOnInexisting is currently set to false on all occasions?

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
```diff
@@ -23,6 +35,42 @@ default HibLanguage create(String languageName, String languageTag, String uuid)
...
 	}
 
+	@Override
+	default void assign(HibLanguage language, HibProject project, EventQueueBatch batch, boolean throwOnExisting) {
+		project.getLanguages().stream()
+			.filter(l -> l.getLanguageTag().equals(language.getLanguageTag()))
+			.findAny()
+			.ifPresentOrElse(existing -> {
+					if (throwOnExisting) {
+						throw error(BAD_REQUEST, "error_language_already_assigned", language.getLanguageTag(), project.getName());
```
npomaroli, line 45: Isn't CONFLICT the better response code here?
Or should we remove this, because throwOnExisting is currently set to false on all occasions?

npomaroli (changes requested):

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
plyhun: I'd rather leave both, since a) NOT_FOUND clearly stands for 'language not found for the context' (and there's hardly an alternative for BAD_REQUEST in such case, except maybe a CONFLICT), b) we don't know at the moment if it gonna be used with true in the future.

mdm/common/src/main/java/com/gentics/mesh/core/data/dao/PersistingLanguageDao.java
plyhun: Same as above. I used to treat BAD_REQUEST as something incorrect coming from the client input, though correctly processed by the server, which fits these cases fairly well.

tests/tests-core/src/main/java/com/gentics/mesh/core/language/LanguageEndpointTest.java
plyhun: I'll just comment those out, as per proposal below.

core/src/main/java/com/gentics/mesh/core/actions/impl/LanguageDAOActionsImpl.java
plyhun: Same for update/delete in PersistingDao.

PR #1579, opened 2024-01-29
SUP-16283: Make bulk request loader follow the limits in the edge cases

plyhun:
## Abstract

There exists a possibility of DDoSing the Elasticsearch server, if a network glitch makes Mesh collecting unlimited ES requests, which are then pushed for the processing all at once.

### General

### On API Changes

### DIFF
--- a/elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
+++ b/elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
@@ -90,7 +90,7 @@ public class BulkOperator implements FlowableOperator<SearchRequest, SearchReque
 				skipIfMultipleThreads(lock, () -> {
 					if (!canceled.get() && requested.get() > 0 && !bulkableRequests.isEmpty() && flushActive.get() && flushing.compareAndSet(true, false)) {
 						timer.stop();
-						log.error("Emitting bulk of size {} to subscriber", bulkableRequests.size());
+						log.debug("Emitting bulk of size {} to subscriber", bulkableRequests.size());
 						do {
 							AtomicLong bulkLength = new AtomicLong(0);
 							List<Bulkable> requests = IntStream.range(0, requestLimit)
--- a/elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
+++ b/elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
@@ -49,6 +49,8 @@ public class BulkOperator implements FlowableOperator<SearchRequest, SearchReque
 	private final long lengthLimit;
 	private ActualBulkOperator<SearchRequest> operator;
 
+	final AtomicBoolean flushActive = new AtomicBoolean(true);
+
 	public BulkOperator(Vertx vertx, Duration bulkTime, int requestLimit, long lengthLimit) {
 		this.vertx = vertx;
 		this.bulkTime = bulkTime.toMillis();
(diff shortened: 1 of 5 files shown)


elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
```diff
@@ -85,12 +89,22 @@ public void drain() {
...
 						log.trace("Emitting bulk of size {} to subscriber", bulkableRequests.size());
-						BulkRequest request = new BulkRequest(bulkableRequests.asList());
-						bulkableRequests.clear();
-						if (log.isDebugEnabled()) {
-							log.debug("Sending bulk to elasticsearch:\n{}", request);
-						}
-						subscriber.onNext(request);
+						do {
+							Long[] bulkLength = new Long[] { 0L };
+							List<Bulkable> requests = IntStream.range(0, requestLimit)
```
plyhun, line 94: Obey number limit.

elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
```diff
@@ -85,12 +89,22 @@ public void drain() {
...
+								.mapToObj(index -> bulkableRequests.poll())
+								.takeWhile(maybeBulkable -> maybeBulkable
+										.filter(bulkable -> bulkLength[0] < lengthLimit)
```
plyhun, line 97: Obey size limit.

tests/tests-core/src/main/java/com/gentics/mesh/search/ElasticSearchProviderTimeoutTest.java
```diff
@@ -59,7 +59,14 @@ public static void setupTimeoutServer() throws IOException {
 			}
 		});
 		server.rxListen().blockingGet();
+	}
 
+	@Test
+	public void testHugeBatch() throws IOException {
```
npomaroli, line 65: What does this test actually prove? There is no assertion in here - apart from that creating 1200 users succeeds, but that is completely unrelated to ES indexing.

elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
```diff
@@ -85,12 +89,22 @@ public void drain() {
...
 						timer.stop();
 						log.trace("Emitting bulk of size {} to subscriber", bulkableRequests.size());
-						BulkRequest request = new BulkRequest(bulkableRequests.asList());
-						bulkableRequests.clear();
-						if (log.isDebugEnabled()) {
-							log.debug("Sending bulk to elasticsearch:\n{}", request);
-						}
-						subscriber.onNext(request);
+						do {
+							Long[] bulkLength = new Long[] { 0L };
```
npomaroli, line 93: Use an AtomicLong here

elasticsearch/src/main/java/com/gentics/mesh/search/verticle/bulk/BulkOperator.java
```diff
@@ -85,12 +89,22 @@ public void drain() {
...
+								.map(Optional::get)
+								.peek(bulkable -> bulkLength[0] += bulkable.bulkLength())
+								.collect(Collectors.toList());
+							BulkRequest request = new BulkRequest(requests);
+							if (log.isDebugEnabled()) {
+								log.debug("Sending bulk to elasticsearch:\n{}", request);
+							}
+							subscriber.onNext(request);
+						} while (!bulkableRequests.isEmpty());
 						BackpressureHelper.produced(requested, 1);
```
npomaroli, line 108: Doesn't this have to be changed as well, or at least moved to the do-while loop?

npomaroli (changes requested):
We definitely should have a test case, which shows the error (without the fix) and proves that the fix works.

tests/tests-core/src/main/java/com/gentics/mesh/search/ElasticSearchProviderTimeoutTest.java
plyhun: It is created during a timeout, so the indexing process should remain stable. Again, the direct use case is not reproducible, due to its spontaneous nature.

plyhun: I'll check if I'm able to extend BulkOperatorTest. Which will still be kinda artificial...

plyhun: Have made it eventually, with an extension that I don't really like.

PR #1580, opened 2024-01-29
Optimization of user/group/role sync

npomaroli:
## Abstract

Synchronization of users, groups and roles from AuthServicePlugin possibly handles multiple elements, in order to improve performance, the implementation is changed to support batch loading where possible.

Also parallel requests might handle the same entities, which can cause race conditions. In order to better handle such situations (without letting the request fail), the sync process is repeated once in case of an error.

### General

### On API Changes

### DIFF
--- a/mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/dao/impl/GroupDaoWrapperImpl.java
+++ b/mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/dao/impl/GroupDaoWrapperImpl.java
@@ -5,6 +5,9 @@ import static com.gentics.mesh.core.data.relationship.GraphRelationships.HAS_ROL
 import static com.gentics.mesh.core.data.relationship.GraphRelationships.HAS_USER;
 import static com.gentics.mesh.core.data.util.HibClassConverter.toGraph;
 
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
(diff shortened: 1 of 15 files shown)


mdm/orientdb-wrapper/src/main/java/com/gentics/mesh/core/data/dao/impl/GroupDaoWrapperImpl.java
```diff
@@ -84,6 +86,17 @@ public Result<? extends HibRole> getRoles(HibGroup group) {
 		return groupRoot.getRoles(graphGroup);
 	}
 
+	@Override
+	public Map<HibGroup, Collection<? extends HibRole>> getRoles(Collection<HibGroup> groups) {
```
plyhun, line 90: This could've been put into PersistingGroupDao as default method, since its does not utilize any DB specific data.

npomaroli: Yes but we don't want that as default implementation, because it's inefficient for Mesh SQL.

plyhun: SQL overrides it anyway.

plyhun (approved):

PR #1586, opened 2024-02-07
Hotfix 2.0.x SUP-16159 Improve WebSocket error handling

yrucrem:
## Abstract

Errors from Vert.x when creating an eventbus WebSocket where not handled which caused a NullPointerException during event creation. Typical errors like invalid event addresses are now handled, and multiple registrations of the same address no longer lead to errors.
### General

### On API Changes


yrucrem:
Will be upmerged from hotfix-1.10.x.

PR #1589, opened 2024-02-12
SUP-16105: Support for newer Keycloak versions

plyhun:
## Abstract

An ability to reformat the Keycloak request URL.

### General

### On API Changes

### DIFF
--- a/services/jwt-auth/src/main/java/com/gentics/mesh/auth/oauth2/MeshOAuth2ServiceImpl.java
+++ b/services/jwt-auth/src/main/java/com/gentics/mesh/auth/oauth2/MeshOAuth2ServiceImpl.java
@@ -42,7 +42,6 @@ import com.gentics.mesh.core.data.user.MeshAuthUser;
 import com.gentics.mesh.core.db.Database;
 import com.gentics.mesh.core.db.Tx;
 import com.gentics.mesh.core.endpoint.admin.LocalConfigApi;
-import com.gentics.mesh.core.rest.error.NameConflictException;
 import com.gentics.mesh.core.rest.group.GroupReference;
 import com.gentics.mesh.core.rest.group.GroupResponse;
 import com.gentics.mesh.core.rest.role.RoleReference;
@@ -429,6 +428,7 @@ public class MeshOAuth2ServiceImpl implements MeshOAuthService {
 	 * @param admin admin user
 	 * @throws CannotWriteException
 	 */
+	@SuppressWarnings({ "unchecked", "rawtypes" })
 	public void handleMappingResult(Tx tx, EventQueueBatch batch, MappingResult result, HibUser authUser, HibUser admin) throws CannotWriteException {
 		RoleDao roleDao = tx.roleDao();
 		GroupDao groupDao = tx.groupDao();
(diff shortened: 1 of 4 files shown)


services/jwt-auth/src/main/java/com/gentics/mesh/auth/oauth2/MeshOAuth2ServiceImpl.java
```diff
@@ -470,7 +470,7 @@ public void handleMappingResult(Tx tx, EventQueueBatch batch, MappingResult resu
 		});
 
 		// check which groups are not yet assigned to the user and add them
-		List<? extends HibGroup> assignedGroups = userDao.getGroups(authUser).list();
+		List<HibGroup> assignedGroups = new ArrayList<>(userDao.getGroups(authUser).list());
```
plyhun, line 473: Here and below - regression of #1580 .

services/jwt-auth/src/main/java/com/gentics/mesh/auth/util/KeycloakUtils.java
```diff
@@ -148,5 +202,4 @@ private static OkHttpClient httpClient() {
 		OkHttpClient client = builder.build();
 		return client;
 	}
```
plyhun, line 204: The ultimate question - why we have something Keycloak-related in Vanilla Mesh anyway, while it should be agnostic of any identity provider?

services/jwt-auth/src/main/java/com/gentics/mesh/auth/util/KeycloakUtils.java
```diff
@@ -61,9 +65,24 @@ public static String loadPublicKey(String realmName, String authServerUrl) {
...
+	 * 
+	 * @param protocol
+	 * @param host
+	 * @param port
+	 * @param realmName
+	 * @param maybeCustomRealmUrlFormat optional realm URL format, suitable for Java formatter with parameters [protocol, host, port, realm name], e.g. '%s://%s:%d/auth/realms/%s'
+	 * @return
+	 * @throws IOException
+	 */
+	public static Set<JsonObject> loadJWKs(String protocol, String host, int port, String realmName, Optional<String> maybeCustomRealmUrlFormat) throws IOException {
```
npomaroli, line 82: I am not entirely happy with the format of custom Keycloak URLs. While it is ok for internal usage, we expose this to a configuration parameter of the keycloak plugin and this may be error-prone and inconvenient, especially for users who are not familiar with java.

plyhun: As an alternative, this may be implemented as following the Keycloak `realms URL prefix` convention. 
Aside of that I'm totally unhappy of having Keycloak references in an unrelated code.

services/jwt-auth/src/main/java/com/gentics/mesh/auth/util/KeycloakUtils.java
plyhun: Most probably because most of the KC-integration tests are baked into Mesh UT suite...

services/jwt-auth/src/main/java/com/gentics/mesh/auth/util/KeycloakUtils.java
```diff
@@ -26,6 +27,11 @@ public final class KeycloakUtils {
 
 	private static final Logger log = LoggerFactory.getLogger(KeycloakUtils.class);
 
+	private static final String DEFAULT_REALM_URL_FORMAT = "%s://%s:%d%s/realms/%s";
```
npomaroli, line 30: This is not used any more, or is it?

npomaroli (approved):

PR #1598, opened 2024-03-13
SUP-14730: Use HttpOnly cookie flag

plyhun:
## Abstract

Potential XSS attacks prevention.

### General

### On API Changes

### DIFF
--- a/common/src/main/java/com/gentics/mesh/auth/handler/MeshJWTAuthHandler.java
+++ b/common/src/main/java/com/gentics/mesh/auth/handler/MeshJWTAuthHandler.java
@@ -149,9 +149,12 @@ public class MeshJWTAuthHandler extends AuthenticationHandlerImpl<MeshJWTAuthPro
 				if (!result.isUsingAPIKey()) {
 					String jwtToken = authProvider.generateToken(authenticatedUser);
 					// Remove the original cookie and set the new one
-					context.removeCookie(SharedKeys.TOKEN_COOKIE_KEY);
-					context.addCookie(Cookie.cookie(SharedKeys.TOKEN_COOKIE_KEY, jwtToken)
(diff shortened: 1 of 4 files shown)


common/src/main/java/com/gentics/mesh/auth/handler/MeshJWTAuthHandler.java
```diff
@@ -149,9 +149,12 @@ private void handleJWTAuth(RoutingContext context, boolean ignoreDecodeErrors) {
 				if (!result.isUsingAPIKey()) {
 					String jwtToken = authProvider.generateToken(authenticatedUser);
 					// Remove the original cookie and set the new one
-					context.removeCookie(SharedKeys.TOKEN_COOKIE_KEY);
-					context.addCookie(Cookie.cookie(SharedKeys.TOKEN_COOKIE_KEY, jwtToken)
-						.setMaxAge(meshOptions.getAuthenticationOptions().getTokenExpirationTime()).setPath("/"));
+					context.response().removeCookie(SharedKeys.TOKEN_COOKIE_KEY);
+					context.response().addCookie(Cookie.cookie(SharedKeys.TOKEN_COOKIE_KEY, jwtToken)
+						.setHttpOnly(true)
```
plyhun, line 154: @yrucrem 
I'm not really sure if it can/should be parameterized. AFAIK our Java REST client ignores this setting. Did not try the PHP/JS clients though. Need an advice here, if possible.

yrucrem: I don't think it's necessary to make this configurable, always enabling httpOnly should not be a problem (portal-java does this as well). As for tests: If there are already integration tests which check if the authentication cookie is set, the flag could be checked there. Otherwise I think manual checks suffice since this is not a feature on our part but Vert.x.

PR #1603, opened 2024-05-02
SUP-16720: Fix for several simultaneous similar requests (Node only)

plyhun:
## Abstract

If several similar, but differently named GraphQL requests are requested within the same context, only the latest is executed.

### General

### On API Changes


verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
```diff
@@ -61,22 +61,27 @@ public class NodeDataLoader {
 	 * Batch load all field containers of a node for transaction branch and the provided types, without doing permission
 	 * checks.
 	 */
-	public static BatchLoaderWithContext<HibNode, List<HibNodeFieldContainer>> CONTENT_LOADER = (keys, environment) -> {
+	public static BatchLoaderWithContext<DataLoaderKey<HibNode>, List<HibNodeFieldContainer>> CONTENT_LOADER = (keys, environment) -> {
```
plyhun, line 64: See the explanation in `DataLoaderKey`.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/DataLoaderKey.java
```diff
@@ -0,0 +1,53 @@
...
+import graphql.schema.DataFetchingEnvironment;
+
+/**
+ * The GraphQL path-aware key for {@link DataLoader}.
+ * 
+ * @param <T>
+ */
+public final class DataLoaderKey<T> {
+
+	private final ResultPath path;
```
plyhun, line 17: So we take a GraphQL path into account, at last. So `data.rootMerged.contentChildren.elements` is separated from `data.rootMerged.folderChildren.elements`, and batch loader treat these two as separate queries, as the keys now differ.
Why this was working for non-native filtering? Because there were no filtering applied at the fetch stage, and so every query was ok with any data coming from the loader, performing the filtering afterwards.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/DataLoaderKey.java
```diff
@@ -0,0 +1,58 @@
...
+	private final String path;
+	private final T value;
+
+	public DataLoaderKey(String path, T value) {
+		this.path = path;
+		this.value = value;
+	}
+
+	public DataLoaderKey(ResultPath path, T value) {
+		this(path.toString().replaceAll("\\[[\\d+]\\]", StringUtils.EMPTY), value);
```
plyhun, line 27: Result indices break the query paths, so cutting them off.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
```diff
@@ -352,12 +373,17 @@ public boolean equals(Object o) {
 			if (this == o) return true;
 			if (!(o instanceof Context)) return false;
 			Context context = (Context) o;
-			return type == context.type && Objects.equals(languageTags, context.languageTags);
+			return type == context.type 
+					&& Objects.equals(languageTags, context.languageTags) 
+					&& Objects.equals(paging.toString(), context.paging.toString())
```
plyhun, line 378: Paging and native filter have to be compared by value, not by instance.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/DataLoaderKey.java
npomaroli: Please add a comment about this in the Javadoc also.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
npomaroli: paging.toString() will return the paging parameters as query string, but as far as I have seen, there is no guarantee about the order of the parameters in the resulting string

npomaroli (commented):
Apart from some more Javadoc and my comment about comparing PagingParameters string representation, the changes look good.

PR #1604, opened 2024-05-02
Replace library for webp handling

npomaroli:
### General

### On API Changes


common/src/main/java/com/gentics/mesh/core/image/spi/AbstractImageManipulator.java
```diff
@@ -125,6 +126,8 @@ public Single<ImageInfo> readImageInfo(String path) {
 		return result.toSingle();
 	}
 
+	abstract protected BufferedImage readFromFile(File imageFile) throws IOException;
```
npomaroli, line 129: Javadoc

services/image-imgscalr/pom.xml
```diff
@@ -17,7 +17,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<imgscalr-lib.version>4.2</imgscalr-lib.version>
 		<twelvemonkeys.version>3.10.1</twelvemonkeys.version>
-		<imageio.webp.version>0.1.6</imageio.webp.version>
+		<scrimage.webp.version>4.1.3</scrimage.webp.version>
```
npomaroli, line 20: Remove property

PR #1611, opened 2024-07-03
GPU-1625: Elasticsearch 8.x compatibility

yrucrem:
## Abstract

Make sure Mesh is compatible with Elasticsearch 8.x. New compliance mode ES_8 added. Documentation updated on how to create HTTPS certificates. Mapping and index operations updated to work with ES 8. Added ES 8 version to test container and changed tests to run with ES 6, ES 7 and ES 8.

### General

* [X ] Added abstract that describes the change
* [X] Ensured that the change is documented in the docs

### On API Changes

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -26,7 +26,7 @@
 		<asm.version>3.3.1</asm.version>
 		<spring.security.version>5.5.7</spring.security.version>
 		<ferma.version>${project.version}</ferma.version>
-		<elasticsearch.client.version>1.2.0-SNAPSHOT</elasticsearch.client.version>
+		<elasticsearch.client.version>1.1.1</elasticsearch.client.version>
 		<orientdb.version>3.2.10</orientdb.version>
 		<hazelcast.version>3.12.13</hazelcast.version>
 		<jackson.version>2.17.0</jackson.version>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -26,7 +26,7 @@
 		<asm.version>3.3.1</asm.version>
 		<spring.security.version>5.5.7</spring.security.version>
 		<ferma.version>${project.version}</ferma.version>
-		<elasticsearch.client.version>1.1.1</elasticsearch.client.version>
+		<elasticsearch.client.version>1.2.0-SNAPSHOT</elasticsearch.client.version>
 		<orientdb.version>3.2.10</orientdb.version>
 		<hazelcast.version>3.12.13</hazelcast.version>
 		<jackson.version>2.17.0</jackson.version>
--- a/elasticsearch/src/main/java/com/gentics/mesh/search/impl/ElasticSearchProvider.java
+++ b/elasticsearch/src/main/java/com/gentics/mesh/search/impl/ElasticSearchProvider.java
@@ -97,8 +97,7 @@ public class ElasticSearchProvider implements SearchProvider {
 
 	@Override
 	public ElasticSearchProvider init() {
-
-        return this;
+		return this;
 	}
 
 	@Override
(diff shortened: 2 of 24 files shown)


elasticsearch/src/main/java/com/gentics/mesh/search/impl/ElasticSearchProvider.java
```diff
@@ -97,7 +97,42 @@ public void start() {
...
+		if (settingsDestructiveRequiresName.isEmpty() || settingsDestructiveRequiresName.get()) {
+			log.info("Updating cluster configuration: destructive_requires_name=false");
+
+			JsonObject update = new JsonObject()
+				.put("persistent", new JsonObject()
+					.put("action", new JsonObject()
+						.put("destructive_requires_name", false)));
+
+            try {
+                settings = client.updateClusterSettings(update).sync();
```
npomaroli, line 125: Updating the elasticsearch configuration like this feels like a bad idea:
What if the client is not allowed to do this? (init of the provider will log an error message and the setting will not be changed, bad things will follow)
What if elasticsearch instance is recreated and Mesh not restarted? (setting is back to the configured state, bad things will follow)
We do not mention anything of that in our documentation. Silently fiddling with our customers servers is a bad habit.

Why not just documenting, how elasticsearch needs to be configured to properly work with Mesh?

tests/tests-core/src/main/java/com/gentics/mesh/search/CustomIndexSettingsTest.java
```diff
@@ -338,10 +338,7 @@ public void testMicroschemaValidationSucess() {
 	}
 
 	private String typeName() {
-		String name = "default";
-		if (complianceMode() == ComplianceMode.ES_7) {
-			name = "_doc";
-		}
-		return name;
+		// Mapping types have been removed in ES 7 and ES 7+ uses "_doc" as dummy.
+        return complianceMode() == ComplianceMode.ES_6 ? "default" : "_doc";
```
npomaroli, line 342: Formatting (use tabs instead of spaces for indentation)

tests/tests-core/src/main/java/com/gentics/mesh/test/context/MeshTestContext.java
```diff
@@ -641,12 +644,26 @@ public MeshOptions init(MeshTestSetting settings) throws Exception {
 		S3Options s3Options = meshOptions.getS3Options();
 		searchOptions.setTimeout(10_000L);
 
-		String version = ElasticsearchContainer.VERSION_ES6;
-		switch (settings.elasticsearch()) {
-		case CONTAINER_ES7:
-			searchOptions.setComplianceMode(ComplianceMode.ES_7);
-			version = ElasticsearchContainer.VERSION_ES7;
+		String version = switch (settings.elasticsearch()) {
+            case CONTAINER_ES7 -> {
```
npomaroli, line 648: Formatting (use tabs instead of spaces for indentation).

bom/pom.xml
```diff
@@ -26,7 +26,7 @@
 		<asm.version>3.3.1</asm.version>
 		<spring.security.version>5.5.7</spring.security.version>
 		<ferma.version>${project.version}</ferma.version>
-		<elasticsearch.client.version>1.1.1</elasticsearch.client.version>
+		<elasticsearch.client.version>1.2.0-SNAPSHOT</elasticsearch.client.version>
```
npomaroli, line 29: We probably do not need that at all (see comment in ElasticSearchProvider)

npomaroli (changes requested):

npomaroli (approved):

PR #1613, opened 2024-08-05
SUP-17105: Document logging & async check better

plyhun:
### General

### On API Changes


mdm/api/src/main/java/com/gentics/mesh/parameter/impl/ConsistencyCheckParametersImpl.java
```diff
@@ -35,7 +35,7 @@ public ConsistencyCheckParametersImpl() {
 		// async
 		QueryParameter asyncParameter = new QueryParameter();
 		asyncParameter.setDefaultValue("false");
-		asyncParameter.setDescription("Specifiy whether the consistency check should be executed asynchronously (as job).");
+		asyncParameter.setDescription("Specifiy whether the consistency check should be executed asynchronously (as job). *Please not that some checks do require this flag set!*");
```
npomaroli, line 38: "note", not "not"

npomaroli (changes requested):

PR #1615, opened 2024-08-27
SUP-17264: Don't jump over several schema versions during migration

plyhun:
## Abstract

It should not be allowed to perform a (micro)node migration over two non-adjacent (micro)schema versions, since there is no way to determine and apply the correct set of changes.

### General

### On API Changes


plyhun:
Open question is: if the chain of versions is not possible to build - what to do?
I come to the only solution of dropping the problematic branch/node completely - would it be a legitimate target for the consistency check/repair step?

PR #1616, opened 2024-08-30
SUP-17282: Fix data loss on a version-skipping node migration edge case

plyhun:
## Abstract

A data loss is possible, when a set of manual rename/re-type schema changes is applied with no immediate migration triggered.

### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/migration/AbstractMigrationHandler.java
```diff
@@ -112,26 +116,35 @@ protected void prepareMigration(HibFieldSchemaVersionElement<?, ?, ?, ?, ?> from
...
+				currentContent.updateAndGet(oldContent -> () -> oldContent.getFields().putAll(currentChanges));
+
 				// If a migration has been started over non-adjacent from/to versions, the intermediate changes need no actual storage, but a validation..
-				FieldSchemaContainerVersion schema = pair.getKey().getNextVersion().getSchema();
+				FieldSchemaContainerVersion schema = pair.getValue().getNextVersion().getSchema();
 				if (schema instanceof FieldSchemaContainerVersion && !((FieldSchemaContainerVersion)schema).getVersion().equals(newContainer.getSchemaContainerVersion().getVersion())) {
 					log.info("Update skipped for container [{}] of schema [{}] version [{}] from the intermediate version [{}]", newContainer.getUuid(), newContainer.getSchemaContainerVersion().getSchema().getName(), newContainer.getSchemaContainerVersion().getVersion(), ((FieldSchemaContainerVersion)schema).getVersion());
 					schema.assertForUnhandledFields(fm);
 				} else {
+					if (versionChain.size() > 1) {
```
plyhun, line 144: No need to iterate over every data change, if there is only one version to upgrade.

npomaroli (approved):

PR #1617, opened 2024-09-05
Fix validation of image resize with and height to calculate the target

npomaroli:
values before checking limits

## Abstract

When checking that the resized image is within the allowed limits, the width/height has to be calculated before (in cases, where at least one value is set to "auto" or not set at all).

### General

### On API Changes


plyhun (approved):

plyhun:
There are breaking changes, which make it impossible to fetch the already existing image variants. Needs refactoring.

PR #1628, opened 2024-10-04
Runtime related refactoring

plyhun:
## Abstract

This PR mostly consists of moving the functions around, and making the interfaces out of stuff, to have it overridden in the alternative implementations.

### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/endpoint/admin/AdminHandler.java
```diff
@@ -219,6 +219,7 @@ public MeshServerInfoModel getMeshServerInfoModel(InternalActionContext ac) {
 			info.setMeshVersion(Mesh.getPlainVersion());
 			info.setVertxVersion(VersionCommand.getVersion());
 			info.setMeshNodeName(options.getNodeName());
+			info.setMeshRevision("OSS");
```
plyhun, line 222: Here we tell the Mesh revisions apart.

PR #1635, opened 2024-11-19
Hotfix 2.1.x sup 17710

plyhun:
## Abstract

Speed up the graphql queries by correctly merging the contexts.

### General

### On API Changes

### DIFF
--- a/verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
+++ b/verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
@@ -337,12 +337,9 @@ public class NodeDataLoader {
 	 * The context of the loader for a specific key
 	 */
 	public static class Context {
-		private final ContainerType type;
-		
+		private final ContainerType type;		
 		private final String languageTags;
-
 		private final Optional<FilterOperation<?>> maybeNativeFilter;
-
 		private final PagingParameters paging;
 		private final int pagingId;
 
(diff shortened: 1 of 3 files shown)


verticles/graphql/src/main/java/com/gentics/mesh/graphql/dataloader/NodeDataLoader.java
```diff
@@ -376,13 +373,13 @@ public boolean equals(Object o) {
 					&& pagingId == context.pagingId
 					&& Objects.equals(languageTags, context.languageTags)
 					&& Objects.equals(
-							maybeNativeFilter.map(Objects::toString).orElse(StringUtils.EMPTY), 
-							context.maybeNativeFilter.map(Objects::toString).orElse(StringUtils.EMPTY));
+							maybeNativeFilter.map(FilterOperation::toSql).orElse(StringUtils.EMPTY),
```
plyhun, line 376: Long story short: when a `NodeDataLoader` fills up the fetch context, it forms a `HashMap`, where the key is formed from this `Context`. The `FilterOperation` is however misbehaving, as must be compared by its SQL representation, and not default `toString()`, so the similar operations will be merged into one context, which reduces the run time and query count.

plyhun: The UT for it is placed in an Enterprise, since it is the one who suffered.

yrucrem (approved):

PR #1642, opened 2024-12-19
SUP 17176: Fix accessing files in the binary image cache to not create folders of

npomaroli:
the old structure.
Fix the image cache migration to delete all folders of the old structure

## Abstract

Accessing the binary image cache must not create directories of the old structure while searching cache files.
The migration of the old structure must delete the old directories even if they are empty or contain cache files which cannot be migrated because the binary no longer exists.

### General

### On API Changes


plyhun:
Shouldn't this get implemented up from 2.1.x?

npomaroli:
> Shouldn't this get implemented up from 2.1.x only?
No because it is currently messed up in 1.10.x (which is still in use).

plyhun (approved):

PR #1648, opened 2025-01-15
SUP-17558: Fix the abandoned binary data, caused by the unstable natu…

plyhun:
…re of BulkActionContext

## Abstract

The current nature of BulkActionContext is weak, allowing it being abandoned at any time with no processing, on the one hand, and ignoring or misplacing it, on the other. Here the part of the BAC refactoring consists of injecting it into the Tx, the same as is done with the event queue, but with paying attention to its processing, started manually (this is the main difference with event queue, that is executed on Tx success automatically). Plus, the extra actions for deletion the folders traces for the deleted binary is implemented, as well as the check/repair for the already abandoned binaries.

### General

### On API Changes


common/src/main/java/com/gentics/mesh/context/impl/BulkActionContextImpl.java
```diff
@@ -50,18 +50,25 @@ public void process() {
...
+
 	@Override
 	public void process(boolean force) {
 		if (elementCounter.incrementAndGet() >= DEFAULT_BATCH_SIZE || force) {
-			log.info("Processing transaction batch {" + batchCounter.get() + "}. I counted {" + elementCounter.get() + "} elements.");
+			log.info("Processing transaction batch {" + batchCounter.get() + "}. I counted {" + elementCounter.get() + "} elements. {" + asyncActions.size() + "} to be executed.");
 			// Check before commit to ensure we are 100% safe
 			db.blockingTopologyLockCheck();
-			Tx.get().commit();
+			Tx.maybeGet().ifPresent(Tx::commit);
```
plyhun, line 65: BAC is allowed to exist outside of the Tx. Every Tx-dependent action is normally executed manually on demand.

core/src/main/java/com/gentics/mesh/core/actions/impl/BranchDAOActionsImpl.java
```diff
@@ -80,7 +80,7 @@ public boolean update(Tx tx, HibBranch branch, InternalActionContext ac, EventQu
 	}
 
 	@Override
-	public void delete(Tx tx, HibBranch element, BulkActionContext bac) {
+	public void delete(Tx tx, HibBranch element) {
```
plyhun, line 83: Here and below -- since we consider using BAC from the Tx, we need the explicit BAC arguments no more.

core/src/main/java/com/gentics/mesh/core/endpoint/admin/consistency/UploadsConsistencyCheck.java
```diff
@@ -0,0 +1,99 @@
...
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
+public class UploadsConsistencyCheck implements ConsistencyCheck {
+
+	private static final Logger log = LoggerFactory.getLogger(UploadsConsistencyCheck.class);
+
+	@Override
+	public ConsistencyCheckResult invoke(Database db, Tx tx1, boolean attemptRepair) {
+		Transactional<ConsistencyCheckResult> action = db.transactional(tx -> {
```
plyhun, line 24: The uploads cleanup. As every consistency check, may start as check or repair.

mdm/api/src/main/java/com/gentics/mesh/core/data/HibFieldContainer.java
```diff
@@ -134,16 +132,7 @@ default HibField getField(String fieldKey) {
...
 	 */
-	void removeField(String fieldKey, BulkActionContext bac);
-
-	/**
-	 * Remove the field with the given key and use a dummy bulk action context.
-	 * 
-	 * @param fieldKey
-	 */
-	default void removeField(String fieldKey) {
-		removeField(fieldKey, new DummyBulkActionContext());
```
plyhun, line 145: All misbehaving calls to `DummyBulkActionContext` were removed as well.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/binary/BinaryFieldTest.java
```diff
@@ -287,6 +292,9 @@ public void testRemoveRequiredFieldViaNull() {
 				updateContainer(ac, container, BINARY_FIELD, null);
 			});
 		}
+		tx(tx -> {
```
npomaroli, line 295: Consistency checks are executed in every test anyways.

tests/tests-core/src/main/java/com/gentics/mesh/core/node/NodeImageResizeEndpointTest.java
```diff
@@ -232,7 +232,7 @@ public void testTransformImage() throws Exception {
 		ImageManipulationParameters params = new ImageManipulationParametersImpl().setWidth(100);
 		NodeResponse transformResponse = call(() -> client().transformNodeBinaryField(PROJECT_NAME, nodeUuid, "en", version, "image", params));
 		assertEquals("The image should have been resized", 100, transformResponse.getFields().getBinaryField("image").getWidth().intValue());
-		MeshCoreAssertion.assertThat(testContext).hasUploads(2, 2).hasTempFiles(0).hasTempUploads(0);
+		MeshCoreAssertion.assertThat(testContext).hasUploads(1, 1).hasTempFiles(0).hasTempUploads(0);
```
npomaroli, line 235: Why did this change?

npomaroli (changes requested):

core/src/main/java/com/gentics/mesh/core/endpoint/admin/consistency/UploadsConsistencyCheck.java
```diff
@@ -0,0 +1,109 @@
...
+			} while (count >= BulkActionContextImpl.DEFAULT_BATCH_SIZE);
+			log.info("Uploads consistency " + (attemptRepair ? "repair" : "check") + " finished successfully. Please clean your tmpdir right away, if required.");
+			return result;
+		});
+		return action.runInExistingTx(tx1);
+	}
+
+	@Override
+	public String getName() {
+		return "uploads";
```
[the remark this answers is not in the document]
plyhun: Ouch.

core/src/main/java/com/gentics/mesh/core/endpoint/node/NodeCrudHandler.java
```diff
@@ -67,18 +67,15 @@ public void handleDelete(InternalActionContext ac, String uuid) {
 		validateParameter(uuid, "uuid");
 
 		try (WriteLock lock = writeLock.lock(ac)) {
-			utils.syncTx(ac, tx -> {
+			utils.syncTxBulkable(ac, (tx, bac) -> {
```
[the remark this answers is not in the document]
plyhun: With `bulkable` we explicitly inject the `BulkActionContext`, which guarantees the execution of its entries. Otherwise the possible bulk action will be attached to the Vert.x event queue, which is always there, but does not guarantee the action to be executed, especially if the action is using non-blocking code, which causes deadlocks within Vert.x.

core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryTransformHandler.java
```diff
@@ -334,33 +334,41 @@ public void handleTransformImage(RoutingContext rc, String uuid, String fieldNam
...
 				if (log.isDebugEnabled()) {
 					log.debug("Moving upload with uuid {} and tempId {} into place", binaryUuid, temporaryId);
 				}
-				return binaryStorage.moveInPlace(binaryUuid, temporaryId).andThen(Single.just(n));
+				return binaryStorage.moveInPlace(binaryUuid, temporaryId).andThen(sn);
 			} else {
-				return Single.just(n);
+				return sn;
+			}
+		}).onErrorResumeNext(e -> {
```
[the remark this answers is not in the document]
plyhun: Perhaps it is valid to purge the temporary upload at this point as well, along with the binary storage item. In any case we do not know what have exactly happened, but we do know for sure, that no new binary will be created in the DB, and so no binary data should pass on, if being copied.

tests/tests-core/src/main/java/com/gentics/mesh/core/node/NodeImageResizeEndpointTest.java
plyhun: As we now process the binary files correctly, the old unused binary is no more there.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/binary/BinaryFieldTest.java
plyhun: This is not a check, but repair. This particular test does not use the REST API for the binary manipulation --> this means that in order to please the every-test consistency check, one has to remove the orphaned binary data manually.

mdm/api/src/main/java/com/gentics/mesh/context/impl/DummyBulkActionContext.java
```diff
@@ -37,4 +37,7 @@ public void setRootCause(ElementType type, String uuid, EventCauseAction action)
 	public void add(Completable action) {
 	}
 
+	@Override
```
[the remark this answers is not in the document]
plyhun: I wanted this to be heard aloud explicitly, yes. I'll move it.

core/src/main/java/com/gentics/mesh/core/endpoint/admin/consistency/UploadsConsistencyCheck.java
```diff
@@ -0,0 +1,109 @@
...
+							if (attemptRepair) {
+								log.info("Binary data is invalid and will be moved to the tmpdir: " + path);
+		
+								String segments = uploadsPath.relativize(path.getParent()).toString();
+								Path tmpSegmentsPath = tmpPath.resolve(segments);
+								try {
+									Files.createDirectories(tmpSegmentsPath);
+									Files.move(path, tmpSegmentsPath.resolve(path.getFileName()));
+									Path parent = path.getParent().toAbsolutePath();
+									while (parent != null && parent.compareTo(uploadsPath) != 0 && Files.list(parent).count() < 1 && Files.deleteIfExists(parent)) {
```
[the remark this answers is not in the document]
plyhun: Set a new option `deleteFolderTrace`, defaulted to `true`. Clients with massive parallel uploads of the same data (do we have any?) may reset it to achieve the old behaviour.

npomaroli (approved):

PR #1655, opened 2025-02-18
GPU-485: Additional JWT options

plyhun:
### General

### On API Changes


deckdom (approved):
While I do approve it, NOP/PK should also have a look I think

PR #1657, opened 2025-03-04
Fix the docu build

plyhun:
## Abstract

Fix reusing the docs from Mesh OSS in the EE builds. Add a docu for the backup/restore.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/endpoint/admin/AdminEndpoint.java
+++ b/core/src/main/java/com/gentics/mesh/core/endpoint/admin/AdminEndpoint.java
@@ -305,7 +305,7 @@ public abstract class AdminEndpoint extends AbstractInternalEndpoint {
 		route.path("/debuginfo");
 		route.method(GET);
 		route.description("Downloads a zip file of various [debug information](/docs/administration-guide/#debuginfo) files.");
-		route.addQueryParameter("include", "Information to include. See the [documentation](/docs/administration-guide/#debuginfo) for possible values.", "-backup,consistencyCheck");
+		route.addQueryParameter("include", "Information to include. See the [documentation](/docs/administration-guide/#debuginfo) for possible usages.", "-log,consistencyCheck");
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/gentics/mesh/core/endpoint/admin/AdminEndpoint.java
```diff
@@ -305,7 +305,7 @@ protected void addDebugInfoHandler() {
 		route.path("/debuginfo");
 		route.method(GET);
 		route.description("Downloads a zip file of various [debug information](/docs/administration-guide/#debuginfo) files.");
-		route.addQueryParameter("include", "Information to include. See the [documentation](/docs/administration-guide/#debuginfo) for possible values.", "-backup,consistencyCheck");
+		route.addQueryParameter("include", "Information to include. See the [documentation](/docs/administration-guide/#debuginfo) for possible usages.", "-log,consistencyCheck");
```
plyhun, line 308: Since we have no more `Reflections`, it is either not to hint for the possible values (which weren't served anyway) or to have a static value list just in the docu (not forgetting to update it, if something changes...).

core/src/main/java/com/gentics/mesh/core/endpoint/admin/debuginfo/DebugInfoHandler.java
```diff
@@ -34,7 +34,7 @@ public class DebugInfoHandler {
 	private final Database db;
 	private static final Logger log = LoggerFactory.getLogger(DebugInfoHandler.class);
 	private final Set<String> defaultExcluded = Stream.of(
-		"consistencyCheck").collect(Collectors.toSet());
+		"consistencyCheck","backup").collect(Collectors.toSet());
```
plyhun, line 37: We still have a `backup` provider around, which does nothing.

npomaroli (approved):

PR #1664, opened 2025-03-18
SUP-18278: S3BinaryFieldEndpointTest#testTransformS3BinarySuccessful

plyhun:
## Abstract

The problem arises when a test is started in the scope of an other database context.

### General

### On API Changes

### DIFF
--- a/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryTransformHandler.java
+++ b/core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryTransformHandler.java
@@ -226,11 +226,12 @@ public class BinaryTransformHandler extends AbstractHandler {
 				.handleS3Resize(options.getS3Options().getBucket(), s3ObjectKey, fileName, parameters)
 				.flatMap(file -> {
 					// The image was stored and hashed. Now we need to load the stored file again and check the image properties
-					Single<ImageInfo> info = imageManipulator.readImageInfo(file.getName());
-					Single<FileProps> fileProps = fs.rxProps(file.getName());
+					String filePath = file.getAbsolutePath();
+					Single<ImageInfo> info = imageManipulator.readImageInfo(filePath);
+					Single<FileProps> fileProps = fs.rxProps(filePath);
 					return Single.zip(info, fileProps, (infoV, props) -> {
 						// Return a POJO which hold all information that is needed to update the field
-						return new TransformationResult(null, props.size(), infoV, file.getName());
+						return new TransformationResult(null, props.size(), infoV, filePath);
 					});
 				})
 				.flatMap(transformationResult ->  Single.just(updateNodeInGraph(ac, s3UploadContext, transformationResult, uuid, languageTag, fieldName, parameters)))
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/gentics/mesh/core/endpoint/node/BinaryTransformHandler.java
```diff
@@ -226,11 +226,12 @@ public void handleS3TransformImage(RoutingContext rc, String uuid, String fieldN
 				.handleS3Resize(options.getS3Options().getBucket(), s3ObjectKey, fileName, parameters)
 				.flatMap(file -> {
 					// The image was stored and hashed. Now we need to load the stored file again and check the image properties
-					Single<ImageInfo> info = imageManipulator.readImageInfo(file.getName());
-					Single<FileProps> fileProps = fs.rxProps(file.getName());
+					String filePath = file.getAbsolutePath();
```
plyhun, line 229: The absolute path still stays in Mesh, and is needed for the further steps of an async processing to pick the cached image from another project scope.

services/image-imgscalr/src/main/java/com/gentics/mesh/image/ImgscalrImageManipulator.java
```diff
@@ -406,7 +405,7 @@ public Completable handleS3CacheResize(String bucketName, String cacheBucketName
 					if (res)
 						return s3BinaryStorage.read(cacheBucketName, cacheS3ObjectKey)
 								.flatMapCompletable(cacheFileInfo -> {
-									if (nonNull(cacheFileInfo) || cacheFileInfo.getBytes().length > 0) {
+									if (nonNull(cacheFileInfo) && cacheFileInfo.getBytes().length > 0) {
```
plyhun, line 408: Lint: `cacheFileInfo` is always null.

services/image-imgscalr/src/main/java/com/gentics/mesh/image/ImgscalrImageManipulator.java
```diff
@@ -434,6 +433,14 @@ public Completable handleS3CacheResize(String bucketName, String cacheBucketName
 				});
 	}
 
+	private String getS3CachePath() {
```
plyhun, line 436: S3 cache violates the overall cache placement rules, so is moved to its own folder.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/binary/BinaryFieldEndpointTest.java
```diff
@@ -567,6 +568,20 @@ public void downloadDeletedBinary() throws IOException {
 			NOT_FOUND, "node_error_binary_data_not_found");
 	}
 
+	@Test
+    public void testTransformBinarySuccessful() throws IOException {
```
plyhun, line 572: This test has been added during the investigation of an original issue. Since I found no existing test case, covering the similar issue for the plain binaries, I left it along.

npomaroli (approved):

PR #1665, opened 2025-03-18
Hotfix 2.1.x sup 18280

plyhun:
## Abstract

Since the introduction of a virus checker, a new field on a binary payload prevents a changed data from being compared to the existing one, as the changed has postponed status by default, which is false for the cases where no virus check is set up. The check status is modified after the basic comparison. 

### General

### On API Changes

### DIFF
--- a/rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/impl/BinaryFieldImpl.java
+++ b/rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/impl/BinaryFieldImpl.java
@@ -66,7 +66,7 @@ public class BinaryFieldImpl implements BinaryField {
 
 	@JsonProperty(required = true)
 	@JsonPropertyDescription("Status of the external binary check (one of ACCEPTED, DENIED or POSTPONED). The binary field is only accessible when the status is ACCEPTED.")
-	private BinaryCheckStatus checkStatus = BinaryCheckStatus.ACCEPTED;
+	private BinaryCheckStatus checkStatus = BinaryCheckStatus.POSTPONED;
 
 	@JsonIgnore
 	private String checkSecret;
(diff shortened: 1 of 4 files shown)


tests/tests-core/src/main/java/com/gentics/mesh/core/field/binary/BinaryFieldEndpointTest.java
```diff
@@ -565,6 +565,8 @@ public void downloadDeletedBinary() throws IOException {
 		call(() -> client().transformNodeBinaryField(PROJECT_NAME, updatedResponse.getUuid(), updatedResponse.getLanguage(), updatedResponse.getVersion(),
 			"binary", new ImageManipulationParametersImpl().setWidth(250)),
 			NOT_FOUND, "node_error_binary_data_not_found");
+
```
plyhun, line 568: Graceful cleanup.

rest-model/src/main/java/com/gentics/mesh/core/rest/node/field/impl/BinaryFieldImpl.java
```diff
@@ -66,7 +66,7 @@ public String getType() {
 
 	@JsonProperty(required = true)
 	@JsonPropertyDescription("Status of the external binary check (one of ACCEPTED, DENIED or POSTPONED). The binary field is only accessible when the status is ACCEPTED.")
-	private BinaryCheckStatus checkStatus = BinaryCheckStatus.POSTPONED;
+	private BinaryCheckStatus checkStatus = BinaryCheckStatus.ACCEPTED;
```
npomaroli, line 69: This is not a fix of the problem, it only changes the situation, where the problem occurs.
The real problem is that the request to change a binary field can contain the "checkStatus" field, which is not changeable (and should not be), but Mesh will compare the sent checkStatus (or the default value) with the current checkStatus of the field and if they are different, will create a new version.
The correct solution should be to ignore the (unmodifiable) field "checkStatus" in the request when comparing the fields.

npomaroli (changes requested):

npomaroli (approved):

PR #1675, opened 2025-04-18
Hotfix 2.1.x sup 18387

plyhun:
## Abstract

Fix the language dependent S3 binaries

### General

### On API Changes

### DIFF
--- a/rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
+++ b/rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
@@ -7,16 +7,13 @@ import static com.gentics.mesh.rest.client.impl.HttpMethod.POST;
 import static com.gentics.mesh.rest.client.impl.HttpMethod.PUT;
 import static com.gentics.mesh.util.URIUtils.encodeSegment;
 
-import java.io.ByteArrayInputStream;
 import java.io.InputStream;
-import java.io.SequenceInputStream;
-import java.io.UnsupportedEncodingException;
 import java.security.InvalidParameterException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.Vector;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -1198,7 +1195,11 @@ public abstract class MeshRestHttpClientImpl extends AbstractMeshRestHttpClient 
 		Objects.requireNonNull(projectName, "projectName must not be null");
 		Objects.requireNonNull(nodeUuid, "nodeUuid must not be null");
 
-		String path = "/" + encodeSegment(projectName) + "/nodes/" + nodeUuid + "/binary/" + fieldKey + getQuery(getConfig(), parameters);
+		String path = "/" + encodeSegment(projectName) + "/nodes/" + nodeUuid + "/binary/" + fieldKey 
+				+ getQuery(getConfig(), Stream.of(
+						// the provided node params are ignored
+						Stream.of(parameters).filter(p -> !NodeParametersImpl.class.isInstance(p)), 
+						Stream.of(new NodeParametersImpl().setLanguages(languageTag))).flatMap(Function.identity()));
 
 		return prepareRequest(GET, path, MeshBinaryResponse.class);
 	}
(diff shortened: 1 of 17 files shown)


rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
```diff
@@ -1198,7 +1195,11 @@ public MeshRequest<MeshBinaryResponse> downloadBinaryField(String projectName, S
 		Objects.requireNonNull(projectName, "projectName must not be null");
 		Objects.requireNonNull(nodeUuid, "nodeUuid must not be null");
 
-		String path = "/" + encodeSegment(projectName) + "/nodes/" + nodeUuid + "/binary/" + fieldKey + getQuery(getConfig(), parameters);
+		String path = "/" + encodeSegment(projectName) + "/nodes/" + nodeUuid + "/binary/" + fieldKey 
+				+ getQuery(getConfig(), Stream.of(
+						// the provided node params are ignored
+						Stream.of(parameters).filter(p -> !NodeParametersImpl.class.isInstance(p)),
```
plyhun, line 1201: The distinct `language` argument was ignored all the time.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/s3binary/S3BinaryFieldEndpointTest.java
```diff
@@ -230,20 +234,92 @@ public void testDownloadBinaryWithParams() {
...
+            assertNotNull(response);
+            try {
+                byte[] downloadBytes = IOUtils.toByteArray(response.getStream());
+                InputStream in = new ByteArrayInputStream(downloadBytes);
+                bufDe = ImageIO.read(in);
+            } catch (IOException ioException) {
+                fail();
+            }
+            assertEquals("image/jpeg", response.getContentType());
+            assertEquals(2000, bufDe.getWidth());
```
plyhun, line 304: Different widths = different images. Just as planned.

tests/tests-core/src/main/java/com/gentics/mesh/core/field/s3binary/S3BinaryFieldEndpointTest.java
```diff
@@ -259,19 +335,71 @@ public void testUploadWorking() {
...
+        S3RestResponse s3RestResponse1 = null;
+        //creating
+        call(() -> client().updateNodeS3BinaryField(
+                PROJECT_NAME,
+                s3binaryNode.getUuid(),
+                FIELD_NAME,
+                UPLOAD_REQUEST
+        ));
+        // some very illegal stuff
+        try (Tx tx = tx()) {
```
plyhun, line 368: We have explicitly disallowed creation S3 binaries the old way, but somehow this has to be tested.

rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
npomaroli: But now we ignore instances of NodeParametersImpl, if they are passed to the function. That's also not good. Especially, since for all other requests, everything which is added as query paramater is passed in instances of ParameterProvider, everything, which is part of the URL path, is passed as distinct function parameter.

npomaroli (changes requested):

rest-client/src/main/java/com/gentics/mesh/rest/client/impl/MeshRestHttpClientImpl.java
plyhun: This will be a breaking change then.

plyhun: Did the best for the current API.

npomaroli: Ok, maybe I did not make myself clear enough.
1. I see no reason, why the general rule for all REST API Methods in the client (query Parameters -> ParameterProvider, path Parameters -> distinct Parameters) should be violated for the binary field methods
2. As a consequence, the fact that the "language" Parameter is ignored is correct, because the optional lang parameter is a query parameter for the endpoint
3. "The case of using comma separated language values is also not applicable to the binary download." is not true, I just tested it by creating a node with a binary field in english and german. Getting the binary with "lang=fr" resulted in a 404 (expected), while "lang=fr,de" and "lang=fr,en" correctly returned the german and english version of the binary.

So what we should do is: Add a comment to the method(s) that the "language" parameter is not used (and languages must be passed via ParameterProvider implementations) and remove the parameter in the next feature release.

npomaroli (changes requested):

npomaroli (approved):

PR #1685, opened 2025-07-08
SUP-18746: Fix the parsing of both ISO-8601 variants

plyhun:
## Abstract

As we already have the date parser, use it.

### General

### On API Changes

### DIFF
--- a/rest-model/src/main/java/com/gentics/mesh/util/DateUtils.java
+++ b/rest-model/src/main/java/com/gentics/mesh/util/DateUtils.java
@@ -22,11 +22,11 @@ public final class DateUtils {
 
 	private static final Logger log = LoggerFactory.getLogger(DateUtils.class);
 
-	static ZoneOffset zoneOffset;
+	public static final ZoneOffset ZONE_OFFSET;
 
 	static {
 		OffsetDateTime odt = OffsetDateTime.now(ZoneId.systemDefault());
-		zoneOffset = odt.getOffset();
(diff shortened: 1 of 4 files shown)


rest-model/src/main/java/com/gentics/mesh/util/DateUtils.java
```diff
@@ -92,7 +92,7 @@ public static Long fromISO8601(String dateString, boolean failOnFormatError) {
 			}
 			try {
 				// We also support date strings which do not include an offset. We apply the system offset in those cases.
-				Long date = LocalDateTime.parse(dateString).toInstant(zoneOffset).toEpochMilli();
+				Long date = LocalDateTime.parse(dateString).toInstant(ZONE_OFFSET).toEpochMilli();
```
plyhun, line 95: Good to have a local Timezone around, if one needs it.

rest-model/src/main/java/com/gentics/mesh/util/DateUtils.java
```diff
@@ -22,11 +22,11 @@ public final class DateUtils {
 
 	private static final Logger log = LoggerFactory.getLogger(DateUtils.class);
 
-	static ZoneOffset zoneOffset;
+	public static final ZoneOffset ZONE_OFFSET;
```
plyhun, line 25: Good to have a local Timezone around, if one needs it.

npomaroli (approved):

PR #1754, opened 2026-03-26
Missing GraphQL filters + tests

plyhun:
### General

### On API Changes


core/src/main/java/com/gentics/mesh/core/data/schema/handler/SchemaComparatorImpl.java
```diff
@@ -47,7 +47,7 @@ public List<SchemaChangeModel> diff(SchemaModel schemaA, SchemaModel schemaB) {
 		compareAndAddSchemaProperty(changes, AUTO_PURGE_FLAG_KEY, schemaA.getAutoPurge(), schemaB.getAutoPurge(), SchemaModel.class);
 
 		// .noIndex
-		compareAndAddSchemaProperty(changes, NO_INDEX_KEY, schemaA.getNoIndex(), schemaB.getNoIndex(), SchemaModel.class, true, true);
+		compareAndAddSchemaProperty(changes, NO_INDEX_KEY, schemaA.getNoIndex(), schemaB.getNoIndex(), SchemaModel.class);
```
plyhun: Same as `autoPurge` flag above, we serve here tri-state value, true, false and null, just for the consistency.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/AbstractTypeProvider.java
```diff
@@ -598,16 +598,29 @@ protected PagingParameters getPagingInfo(DataFetchingEnvironment env) {
 			parameters.setPerPage(perPage);
 		}
 		String sortBy = env.getArgument("sortBy");
-		SortOrder sortOrder = env.getArgument("sortOrder");
+		Sorting sortOrder = env.getArgument("sortOrder");
```
plyhun, line 601: Wrong type cast. `Sorting` comes from the `graphql-filter` library.

verticles/graphql/src/main/java/com/gentics/mesh/graphql/type/TagFamilyTypeProvider.java
```diff
@@ -65,9 +64,8 @@ public GraphQLObjectType createType() {
 			GraphQLContext gc = env.getContext();
 			HibTagFamily tagFamily = env.getSource();
 			PagingParameters pagingInfo = getPagingInfo(env);
-			TagFamilyDao tagFamilyDao = Tx.get().tagFamilyDao();
 
-			return tagFamilyDao.getTags(tagFamily, gc.getUser(), pagingInfo);
+			return Tx.get().tagDao().findAll(tagFamily, gc, pagingInfo);
```
plyhun, line 68: Fix for the sorting of the tags of the fetched family. The tag DAO method makes native fetch, with all its possibilities.

PR #1761, opened 2026-04-23
Hotfix 3.1.x sup 19851

plyhun:
### General

### On API Changes


tests/tests-core/src/main/java/com/gentics/mesh/core/field/micronode/ReferencedByMicronodeTest.java
```diff
@@ -0,0 +1,280 @@
...
+import com.gentics.mesh.core.rest.schema.MicronodeFieldSchema;
+import com.gentics.mesh.core.rest.schema.impl.BooleanFieldSchemaImpl;
+import com.gentics.mesh.core.rest.schema.impl.ListFieldSchemaImpl;
+import com.gentics.mesh.core.rest.schema.impl.MicronodeFieldSchemaImpl;
+import com.gentics.mesh.core.rest.schema.impl.NodeFieldSchemaImpl;
+import com.gentics.mesh.test.MeshTestSetting;
+import com.gentics.mesh.util.StreamUtil;
+
+@MeshTestSetting(testSize = FULL, startServer = true)
+public class ReferencedByMicronodeTest extends AbstractFieldTestBase<MicronodeFieldSchema> {
```
plyhun: Having a Micronode referenced more that once from different content types looks legitimate to me. Though it is tricky to achieve. This test does not cover the client case, which has been proven on the original client data.

npomaroli (approved):

PR #1765, opened 2026-06-11
SUP-19994: Fix for the filter of the filename of the binary field of the node list item

plyhun:
### General

### On API Changes


plyhun:
Needs unit test & changelog

Issue #1773, opened 2026-07-02
Permanent worker-thread hang on node update with micronode-list field


kaiminus123:
Hello everyone

## Summary

Updating a node that contains a **micronode-list field** can **permanently block a Vert.x worker thread** when the underlying persistence throws (e.g. a MariaDB deadlock). The exception is swallowed by an RxJava 2 anti-pattern, so the `blockingGet()` in `RestUpdaters.MICRONODE_LIST_UPDATER` never returns. The thread is lost for good; repeated occurrences exhaust the worker pool and the instance stops serving requests.

We observed worker threads blocked for **10–14 hours** in production (a permanent hang, not slowness). This became frequent after migrating to the v3 Hibernate/MariaDB backend, which raises real SQL exceptions where the legacy graph backend did not.

## Environment

- Gentics Mesh **3.2.x** (Hibernate/MariaDB backend)
- Vert.x 5.x, RxJava **2.2.21**
- MariaDB 12.x

## Symptoms

`vertx-blocked-thread-checker` reports threads blocked far beyond the limit, always on the same stack:

```
WARN [vertx-blocked-thread-checker] Thread vert.x-worker-thread-19 has been blocked for 36885865 ms, time limit is 60000 ms
io.vertx.core.VertxException: Thread blocked
    at java.util.concurrent.CountDownLatch.await(...)
    at io.reactivex.Single.blockingGet(Single.java:2870)
    at ...RestUpdaters.lambda$static$11(RestUpdaters.java:509)   // MICRONODE_LIST_UPDATER
    at ...HibUnmanagedFieldContainer.updateFieldsFromRest(...)
    at ...PersistingNodeDao.update(...)
    at ...NodeEndpoint.lambda$addUpdateHandler$23(NodeEndpoint.java:590)
```

The thread is parked on `Single.blockingGet` → `CountDownLatch.await`, i.e. it is waiting for a reactive result that will never arrive.

## Root cause

`HibMicronodeFieldList.update(InternalActionContext, MicronodeFieldList)` uses `Observable.create(...)` and subscribes with a `(onSuccess, onError)` pair. The persistence writes (`removeAll()`, `insertReferenced()`, `deleteReferenced()`) run **inside the `onSuccess` consumer**:

```java
return Observable.<Boolean>create(subscriber -> {
    ...
    ...toList().subscribe(micronodeList -> {          // onSuccess consumer
        removeAll();                                  // Hibernate write — can throw
[...]

PR #1774, opened 2026-07-03
SUP-20057: Refactor micronode list updater to deliver the possible errors and not hog the thread, leading to DDoS

plyhun:
## Abstract

Targets #1773 1773

### General

### On API Changes


mdm/api/src/main/java/com/gentics/mesh/core/data/node/field/list/HibMicronodeFieldList.java
```diff
@@ -78,75 +83,68 @@ default Single<Boolean> update(InternalActionContext ac, MicronodeFieldList list
 			return a;
 		}));
 
-		return Observable.<Boolean>create(subscriber -> {
-
-			Iterator<MicronodeField> it = list.getItems().stream().map(item -> {
-				if (item == null) {
```
plyhun, line 84: Unfortunately, this has to stay, because `Observable.toIterable()` does not like null items and spits its very own error, which will resolve as HTTP 500.

